PR #31, opened 2019-05-14
Add asyncio version of run_query

trxcllnt:
* Splits the `PyConnector` methods into request/response handlers
* Adds a wrapper for each `PyConnector` method that checks whether it's being called from an async function
* Implement just enough functions to make `run_query` work both sync and async:

```python
def run_query_sync(*args, **kwargs):
    return pyblazing.run_query(*args, **kwargs)

async def run_query_async(*args, **kwargs):
    return await pyblazing.run_query(*args, **kwargs)

sync_handle = run_query_sync('SELECT * FROM main.df', {'df': df})
print(sync_handle.columns)

async_handle = await run_query_async('SELECT * FROM main.df', {'df': df})
print(async_handle.columns)
```

cc: @aucahuasi

### DIFF
--- a/pyblazing/__init__.py
+++ b/pyblazing/__init__.py
@@ -15,6 +15,7 @@ from .api import create_table
 from .api import ResultSetHandle
 from .api import _get_client
 
+from .connector import PyConnector
 from .connector.utils import is_caller_async
 from .api import run_query as run_query_sync
 from .aio import run_query as run_query_async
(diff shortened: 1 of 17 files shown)


aucahuasi:
Closed, too old and the architecture has changed since orchestrator is not needed anymore (no more tcp connections to make it asyc) Thanks!

PR #48, opened 2019-08-02
[REVIEW] check authentication token

aucahuasi:


aucahuasi:
Closed, too old and the architecture has changed since orchestrator is not needed anymore

PR #81, opened 2019-10-02
[REVIEW] Feature/robust service start

williamBlazing:
This PR is to make it so that services can all start at the same time and they use a pinging mechanism to know they are all available to use. It also introduces a shutdown command to more elegantly shut down all services

WORTH NOTING (potentially controversial, but I find really useful for devs and users) :
- This exposes two new functions to BlazingContext()
   - ready() : it returns true if all services return a ping. Can take an optional parameter `wait` to have the function call wait untill all services are up. This will be very handy for testing framework, and potentially other customer workflows, error recovery and more     
   - shutdown()


kharoc:
**READY TO REVIEW!**
Single Node: OK
Distributed Mode: OK

PR #86, opened 2019-10-03
[TESTING] added BlazingContext ability to provide a folder to put process logs

williamBlazing:


williamBlazing:
I decided to leave it off by default. Its simple to leave it enabled by default with a default folder and you dont need to tell conda to create the folder. It already checks if the folder exists and if it does not, it will create it for you. You can also give it a relative path and it will assume its relative to $CONDA_PREFIX or you can give it an absolute path

felipeblazing:
> I decided to leave it off by default. Its simple to leave it enabled by default with a default folder and you dont need to tell conda to create the folder. It already checks if the folder exists and if it does not, it will create it for you. You can also give it a relative path and it will assume its relative to $CONDA_PREFIX or you can give it an absolute path

ok im fine with that

PR #129, opened 2019-11-11
[REVIEW] set default channel blazingsql/label/main

mario21ic:
This fix is to avoid the python package blazingsql as a conda channel


GPUtester:

PR #210, opened 2020-01-08
Feature/allocator parameters

### LINKED ISSUE #209: Enable setting cudf memory allocator options in BlazingContext creation
Enable setting cudf memory allocator options in BlazingContext creation

williamBlazing:
This PR allows to set memory allocation parameters in the generation of the BlazingContext()

Resolves #209


williamBlazing:
usage would be something like:
`allocator_options = blazingsql.AllocatorOptions()
allocator_options.initial_pool_size = 1000000
bc = BlazingContext(allocator_options=allocator_options)`

roaramburu:
@williamBlazing is this an optional parameter? Does it default to something if you don't set this?

PR #217, opened 2020-01-09
[REVIEW] Feature/interops migration

### LINKED ISSUE #213: Migrate Interops to new cudf 0.12

jeanp413:
Closes #213


engine/src/LogicalFilter.cpp
```diff
@@ -348,7 +348,7 @@ void add_expression_to_plan(blazing_frame & inputs,
 					// get_scalar_from_string(right_operand,inputs.get_column(right_index).dtype());
 					//					right_scalars.push_back(left);
 
-					left_inputs.push_back(SCALAR_INDEX);  //
+					// left_inputs.push_back(SCALAR_INDEX);  //
```
williamBlazing: why are these commented out?

jeanp413: just so the compiler doesn't complain, it doesn't really matter because this file is deleted in the filter migration PR

Issue #232, opened 2020-01-13
Skip data filtering with Hive Partitioning


williamBlazing:
**----For BlazingSQL Developers----**
Need to make sure that skip-data partitioning works with Hive Metastore

**How and where should this be implemented?**
What part of the code should be feature be implemented? What should the APIs and/or classes look like?

When a Parquet table is being created, the skip-data metadata is collected. That skip-data metadata collection should be able to use the hive metadata cursor to create metadata for the columns that have been partitioned by. Partitioned columns only have one value per partition, which for skip-data metadata, that means that both the min and the max would have that value.

**Other design considerations **
What components of the engine could be affected by this? What functions should we make sure we use/reuse?

**Testing considerations?**
What sort of unit tests and/or End to End tests be implemented to test this?

PR #241, opened 2020-01-15
[TESTING] Skip data filtering with Hive Partitioning

aocsa:
Issue #232


engine/tests/skipdata/expression_tree_test.cpp
```diff
@@ -17,19 +17,23 @@ struct ExpressionTreeTest : public ::testing::Test {
...
-      auto solution =  tree.prefix();
-      std::cout << "solution:\n";
-      std::cout << solution << "\n";
-      EXPECT_EQ(solution, expected);
-    }else {
-      EXPECT_EQ(valid_expr, false);
+    try {
+      expression_tree tree;
+      if (tree.build(prefix)) {
+        std::cout << "before:\n";
```
felipeblazing: move thoese couts

felipeblazing: i mena remove them

PR #248, opened 2020-01-17
[REVIEW] migrate groupby to cudf 0.12

aucahuasi:


engine/src/operators/GroupBy.cpp
```diff
@@ -207,91 +150,82 @@ void distributed_groupby_without_aggregations(
...
-//		aggrgation_output_columns[i].create_gdf_column(aggrgation_output_table.get_column(i));
-//		aggrgation_output_columns[i].set_name(output_column_names[i]);
-	}
+	   cudf::table_view keys {{group_by_columns}};
+	   cudf::experimental::groupby::groupby obj(keys);
+	   
+	   std::pair<std::unique_ptr<cudf::experimental::table>, std::vector<cudf::experimental::groupby::aggregation_result>> output = obj.aggregate( requests );
+
+	   if (output.first->num_columns()>0) {
+		   std::vector< std::unique_ptr<cudf::column> > gby_cols = output.first->release();
```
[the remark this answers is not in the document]
felipeblazing: unless im somehow reading this wrong this is calling obj.aggregate(requests) every single time that it adds and aggregation op, so on somethign like select sum(a), sum(b) , sum(c) it would calcluate

sum(a) then throw away result
both sum(a) then sum(b) then throw away result
now all 3 
throw away all execpt for the first output

engine/src/operators/GroupBy.cpp
```diff
@@ -207,91 +150,82 @@ void distributed_groupby_without_aggregations(
...
+	std::vector< std::unique_ptr<cudf::column> > & output_columns,
 	const std::vector<cudf::type_id> & output_types,
 	const std::vector<std::string> & output_column_names) {
 	for(size_t i = 0; i < agg_ops.size(); i++) {
 		switch(agg_ops[i]) {
-		case GDF_SUM:
-		case GDF_MIN:
-		case GDF_MAX:
-			if(aggregation_inputs[i].get_gdf_column()->size() == 0) {
+		case cudf::experimental::aggregation::Kind::SUM:
```
felipeblazing: is this not implemented it doesnt seem to actually do anything

engine/src/operators/GroupBy.cpp
```diff
@@ -508,19 +462,18 @@ void aggregationsMerger(std::vector<ral::distribution::NodeColumns> & aggregatio
 }
 
 
-void single_node_aggregations(blazing_frame & input,
+std::unique_ptr<ral::frame::BlazingTable> single_node_aggregations(const ral::frame::BlazingTableView & table,
```
felipeblazing: this should not be a function distributed and single node should have the same flow, they just get prepared differently

engine/src/operators/GroupBy.cpp
```diff
@@ -559,15 +512,17 @@ void distributed_aggregations_with_groupby(Context & queryContext,
...
-			std::vector<gdf_column_cpp> result = compute_aggregations(input,
-				group_column_indices,
-				aggregation_types,
-				aggregation_input_expressions,
-				aggregation_column_assigned_aliases);
-			Library::Logging::Logger().logInfo(
-				timer.logDuration(queryContext, "distributed_aggregations_with_groupby async compute_aggregations"));
-			timer.reset();
-			return result;
+
```
felipeblazing: so this is not implemented yet?

engine/src/operators/GroupBy.cpp
```diff
@@ -649,50 +605,53 @@ void distributed_aggregations_without_groupby(Context & queryContext,
...
-		aggregationsMerger(partitionsToMerge, groupColumnIndices, aggregation_types, input);
-		Library::Logging::Logger().logInfo(timer.logDuration(
-			queryContext, "distributed_aggregations_without_groupby part 2 collectPartitions aggregationsMerger"));
-		timer.reset();
-	} else {
-		std::vector<gdf_column_cpp> empty_output_table(aggregatedTable.size());
-		for(size_t i = 0; i < empty_output_table.size(); i++) {
-			empty_output_table[i].create_empty(aggregatedTable[i].get_gdf_column()->type().id(), aggregatedTable[i].name());
-		}
+	// TODO percy william alex port distribution
```
felipeblazing: put the single node version in here, have an if statement at the beginning to see if you need to do the distributed pre steps like distributing and gathering

engine/src/operators/GroupBy.cpp
```diff
@@ -732,37 +691,54 @@ void process_aggregate(blazing_frame & input, std::string query_part, Context *
 		}
 	}
 
+	std::unique_ptr<ral::frame::BlazingTable> ret;
+	
 	if(aggregation_types.size() == 0) {
 		if(!queryContext || queryContext->getTotalNodes() <= 1) {
-			single_node_groupby_without_aggregations(input, group_column_indices);
+			ret = single_node_groupby_without_aggregations(table, group_column_indices);
```
felipeblazing: only have one version that handles both, handle that logic inside that function by passing it the context

engine/src/execution_graph/logic_controllers/LogicalAggregation.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+
+  std::unique_ptr<ral::frame::BlazingTable> computeAggregation(
+    const ral::frame::BlazingTableView & table,
+    std::vector<int> key_indeces,  // if size > 0 that means we are doing a group by and key_indices and table can create keys table_view
+    std::vector<string> agg_value_expressions,
+    std::vector<cudf::experimental::aggregation::Kind> aggs,  // this we use to create aggregation requests or map them to reduction_op for reduce
+    std::vector<std::string> output_column_names)
+
+   internally, this funciton needs to see if its:
```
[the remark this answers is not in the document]
aucahuasi: not crazy, is just that LogicalAggregation.cpp is not inside the cmake project is just a template to describe the group by port

engine/src/operators/GroupBy.cpp
```diff
@@ -171,11 +113,12 @@ void distributed_groupby_without_aggregations(
 		std::launch::async,
 		[](Context & queryContext, std::vector<gdf_column_cpp> & input, const std::vector<int> & group_column_indices) {
 			static CodeTimer timer2;
-			std::vector<gdf_column_cpp> result = groupby_without_aggregations(input, group_column_indices);
-			Library::Logging::Logger().logInfo(timer.logDuration(
-				queryContext, "distributed_groupby_without_aggregations part 1 async groupby_without_aggregations"));
-			timer.reset();
-			return result;
+			// TODO percy william alex port distribution
```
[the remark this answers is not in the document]
aucahuasi: Not yet, yeah we need to unify both flows, but that is not the goal of this PR, we'll address this the next PR with @williamBlazing

engine/src/operators/GroupBy.cpp
aucahuasi: yes it has a TODO there

PR #253, opened 2020-01-20
[REVIEW] Port evaluate query to cudf0.12

aucahuasi:
Related to https://github.com/BlazingDB/blazingsql/issues/244


engine/src/CalciteInterpreter.cpp
```diff
@@ -605,8 +631,9 @@ blazing_frame evaluate_split_query(std::vector<ral::io::data_loader> input_loade
 		// process yourself and return
 
 		if(is_scan(query[0])) {
-			blazing_frame scan_frame;
-			std::vector<gdf_column_cpp> input_table;
+			std::vector< std::unique_ptr<cudf::column> > scan_frame_columns;
```
felipeblazing: it woulds be much better to use a BlazingTable here that is a better wrapper.

engine/src/CalciteInterpreter.cpp
```diff
@@ -605,8 +631,9 @@ blazing_frame evaluate_split_query(std::vector<ral::io::data_loader> input_loade
...
+			std::vector< std::unique_ptr<cudf::column> > input_table;
+			std::vector<std::string> col_names;
```
felipeblazing: use a BlazingTable dont make a seperate container for col_names

engine/src/CalciteInterpreter.cpp
```diff
@@ -644,15 +672,25 @@ blazing_frame evaluate_split_query(std::vector<ral::io::data_loader> input_loade
 				blazing_timer.reset();
 
 				if(is_filtered_bindable_scan(query[0])) {
-					scan_frame.add_table(input_table);
-					// process_filter(queryContext, scan_frame, query[0]);
+					for (int i = 0; i < input_table.size(); ++i) {
```
felipeblazing: input_table is declared but has no size at declaration and nothing is added to it. This will never happen

engine/src/CalciteInterpreter.cpp
```diff
@@ -665,9 +703,15 @@ blazing_frame evaluate_split_query(std::vector<ral::io::data_loader> input_loade
 
 
 			// EnumerableTableScan(table=[[hr, joiner]])
-			scan_frame.add_table(input_table);
+			// TODO percy cudf0.12 could be some memory issues here if input_tables was moved before
+			for (int i = 0; i < input_table.size(); ++i) {
+				scan_frame_columns.push_back(std::move(input_table[i]));
```
felipeblazing: What about gdf inputs. You cant move them.

engine/src/CalciteInterpreter.cpp
```diff
@@ -686,40 +730,51 @@ blazing_frame evaluate_split_query(std::vector<ral::io::data_loader> input_loade
 			}
 		}
 		// these shoudl be split up and run on different threads
-		blazing_frame left_frame;
-		left_frame = evaluate_split_query(input_loaders,
+		std::unique_ptr<ral::frame::BlazingTable> left_frame = evaluate_split_query(input_loaders,
```
felipeblazing: put these on different threads. They used to be on different threads I have no clue when that went away.

engine/src/CalciteInterpreter.cpp
```diff
@@ -686,40 +730,51 @@ blazing_frame evaluate_split_query(std::vector<ral::io::data_loader> input_loade
...
 			split_inequality_join_into_join_and_filter(query[0], new_join_statement, filter_statement);
-			result_frame = ral::operators::process_join(queryContext, left_frame, new_join_statement);
+			//result_frame = ral::operators::process_join(queryContext, left_frame, new_join_statement);
+			
+			const ral::frame::BlazingTableView table_left = left_frame->toBlazingTableView();
+			const ral::frame::BlazingTableView table_right = right_frame->toBlazingTableView();
+			const std::string expression = new_join_statement;
+			
+			//result_frame = ral::operators::process_join(queryContext, left_frame, new_join_statement);
+			std::unique_ptr<ral::frame::BlazingTable> result_frame = ral::processor::processJoin(table_left, table_right, expression);
```
felipeblazing: This is rhte non distributed version. Is that the one it is supposed to call? I understand if thats coming in later

engine/src/CalciteInterpreter.cpp
```diff
@@ -887,10 +973,10 @@ blazing_frame evaluate_query(
 						new_strings = NVStrings::create_from_array(nullptr, 0);
```
felipeblazing: Can you get rid of the NVStrings stuff? We dont need it anymore

engine/src/CalciteInterpreter.cpp
williamBlazing: we are waiting for you to merge that joins stuff

engine/src/CalciteInterpreter.cpp
williamBlazing: that never has run on different threads. it would be a good idea, but lets leave that for bigger than gpu

engine/src/CalciteInterpreter.cpp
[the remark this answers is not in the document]
aucahuasi: fixed thanks!

aucahuasi: yep that one was fixed too thanks

engine/src/CalciteInterpreter.cpp
aucahuasi: was a logic error that was fixed too thanks

engine/src/CalciteInterpreter.cpp
aucahuasi: same

engine/src/CalciteInterpreter.cpp
aucahuasi: thanks

engine/src/CalciteInterpreter.cpp
aucahuasi: was deleted

engine/src/CalciteInterpreter.cpp
aucahuasi: yep fixed

engine/src/CalciteInterpreter.cpp
aucahuasi: good idea we will do that for bigger than gpu

PR #254, opened 2020-01-20
[REVIEW] finishing group by including distribution

williamBlazing:


engine/src/from_cudf/cpp_src/bitmask/null_mask.cu
```diff
@@ -0,0 +1,514 @@
+/*
```
felipeblazing: oh god do we really need this?

engine/src/CalciteExpressionParsing.cpp
```diff
@@ -493,7 +519,24 @@ cudf::experimental::aggregation::Kind get_aggregation_operation(std::string oper
 //	}
 
 	throw std::runtime_error(
-		"In get_aggregation_operation function: aggregation type not supported, " + operator_string);
+		"In get_aggregation_operation_for_groupby function: aggregation type not supported, " + operator_string);
+}
+
+cudf::experimental::reduction_op get_aggregation_operation_for_reduce(std::string operator_string) {
```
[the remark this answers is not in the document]
williamBlazing: This returns a reduction type (as in SUM, MIN, MAX) not a datatype which is what `get_aggregation_output_type` does. And COUNT is missing because the cudf reduction operator does not support COUNT. I am guessing because count is just `size() - null_count()`

engine/src/from_cudf/cpp_src/bitmask/null_mask.cu
williamBlazing: This is just utility code i copied from cudf. We need some of the utility code and the utility code has dependencies. It was easier for me to copy the whole folder.

PR #261, opened 2020-01-21
[REVIEW] feature/cy-blaz-table-cudf012

### LINKED ISSUE #260: Ctyhon Migration to cudf 0.12: show dataframe in python
**Describe the bug**

Please provide the simplest and most complete steps or code that will allow us to reproduce the bug. These should be:
Minimal – Use as little code as possible that still produces the same problem
Complete – Provide all parts needed to reproduce the problem
Verifiable – Test the code you’re about to provide to make sure it reproduces the problem

**Expected behavior**

 - Environment location: [Bare-metal, Docker, Cloud(specify cloud provider)]
 - Method of BlazingSQL install: [conda, Docker, or from source]
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**
Please run and paste the output of the `print_env.sh` script here, to gather any other relevant environment details

**Additional context**

**----For BlazingSQL Developers----**
**Suspected source of the issue**
Where and what are potential sources of the issue

**Other design considerations**
What components of the engine could be affected by this?

gcca:
Fixes #260


GPUtester:

engine/src/cython/engine.cpp
```diff
@@ -138,10 +138,10 @@ ResultSet runQuery(int32_t masterIndex,
 		// Execute query
 
 		std::unique_ptr<ral::frame::BlazingTable> frame = evaluate_query(input_loaders, schemas, tableNames, query, accessToken, queryContext);
-		
+
 		// TODO percy william cudf0.12
 		//make_sure_output_is_not_input_gdf(frame, tableSchemas, fileTypes);
-		
+
 		std::vector<cudf::column_view> columns;
```
williamBlazing: esto parece que no se esta usando. Borralo, por favor

engine/include/engine/engine.h
```diff
@@ -8,12 +8,34 @@
...
+	ResultSet() {
+		columns = {};
+		names = {};
+		blazingTable = nullptr;
+	}
+
+	ResultSet(std::unique_ptr<ral::frame::BlazingTable> blazingTableValue)
+		: blazingTable{std::forward<std::unique_ptr<ral::frame::BlazingTable>>(blazingTableValue)} {}
+
+	ResultSet(const ResultSet & other) {
```
felipeblazing: this is a copy constructor. If you are trying to define a move constructor then you should use ```ResultSet(const ResultSet && other)```

engine/include/engine/engine.h
```diff
@@ -8,12 +8,34 @@
...
+		columns = std::move(other.columns);
+		names = std::move(other.names);
+		blazingTable.swap(const_cast<ResultSet &>(other).blazingTable);
+	}
+
+	ResultSet & operator=(const ResultSet & other) {
```
felipeblazing: this is copy assignment operator not a move operator.  use ```operator = (const ResultSet && other)``` if you are tryign to define a move operator

engine/include/engine/engine.h
```diff
@@ -8,12 +8,34 @@
...
-	ral::frame::BlazingTableView *blazingTableView;
+	std::unique_ptr<ral::frame::BlazingTable> blazingTable;
+
+	ResultSet() {
+		columns = {};
+		names = {};
+		blazingTable = nullptr;
+	}
+
+	ResultSet(std::unique_ptr<ral::frame::BlazingTable> blazingTableValue)
```
felipeblazing: in other parts of the code we are writing this like so 

```
BlazingTable::BlazingTable(
  std::unique_ptr<CudfTable> table,
  std::vector<std::string> columnNames)
  : table(std::move(table)), columnNames(columnNames){

}
```
 so change the forward to just use the move. this allows you to also make declaration simpler.

felipeblazing:
> Can one of the admins verify this patch?

@GPUtester it needs a few changes. Shouldnt take very long after I talk to @gcca

engine/include/engine/engine.h
gcca: I'm adding the copy constructor beacuse it's a restriction of cython. Internally, cython generates cpp code that uses copy constructor and assign operator.

engine/include/engine/engine.h
gcca: I'm adding the assign operator beacuse it's a restriction of cython. Internally, cython generates cpp code that uses copy constructor and assign operator.

PR #266, opened 2020-01-21
[REVIEW] Replacing old context API

rommelDB:
This PR is the final port for evaluate_query, here we use the new experimental::Context and experimental::stuff and invoke the new ral compute primitives based on cudf 0.12


engine/src/CalciteInterpreter.cpp
```diff
@@ -910,7 +909,7 @@ std::unique_ptr<ral::frame::BlazingTable> evaluate_split_query(std::vector<ral::
 //			}
 //		}
 
-//		ral::communication::network::Server::getInstance().deregisterContext(queryContext.getContextToken());
+//		ral::communication::network::experimental::Server::getInstance().deregisterContext(queryContext.getContextToken());
```
felipeblazing: do we no longer need to deregister?

engine/src/distribution/primitives.cpp
```diff
@@ -1108,8 +150,11 @@ void sendSamplesToMaster(Context * context, const BlazingTableView & samples, st
 	auto message =
 		Factory::createSampleToNodeMaster(message_id, context_token, self_node, table_total_rows, samples);
 
+	// // Send message to master
```
felipeblazing: is this in another pr?

engine/src/testing-libgdf.cu
```diff
@@ -219,8 +219,8 @@ static result_pair systemCommand(uint64_t accessToken, Buffer&& requestPayloadBu
 
       std::thread([]() {
           std::this_thread::sleep_for(std::chrono::seconds(1));
-          ral::communication::network::Client::closeConnections();
-          ral::communication::network::Server::getInstance().close();      
+          ral::communication::network::experimental::Client::closeConnections();
```
felipeblazing: This entire file is no longer used it can be deleted right?

engine/src/CalciteInterpreter.cpp
aucahuasi: Don't know I think this was in the legacy code maybe @williamBlazing can answer this

engine/src/distribution/primitives.cpp
aucahuasi: Yep the distribution part seems incomplete so I guess will be addressed in another one

engine/src/testing-libgdf.cu
aucahuasi: good catch ... deleted!

engine/src/CalciteInterpreter.cpp
williamBlazing: we do want to deregister context

PR #268, opened 2020-01-22
[REVIEW] Refactoring TableSchema flow in cython

### LINKED ISSUE #264: [BUG]  Migration of TableSchema in the cython layer
**Describe the bug**
TableSchema now should have a new attribute:
`ral::frame::BlazingTableView blazingTableView;` 
instead of:
```
std::vector<cudf::column *> columns;
std::vector<std::string> names;
```

Please provide the simplest and most complete steps or code that will allow us to reproduce the bug. These should be:
Minimal – Use as little code as possible that still produces the same problem
Complete – Provide all parts needed to reproduce the problem
Verifiable – Test the code you’re about to provide to make sure it reproduces the problem

**Expected behavior**

 - Environment location: [Bare-metal, Docker, Cloud(specify cloud provider)]
 - Method of BlazingSQL install: [conda, Docker, or from source]
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**
Please run and paste the output of the `print_env.sh` script here, to gather any other relevant environment details

**Additional context**

**----For BlazingSQL Developers----**
**Suspected source of the issue**
Where and what are potential sources of the issue

**Other design considerations**
What components of the engine could be affected by this?

Christian8491:
close #264


GPUtester:

engine/bsql_engine/io/io.pyx
```diff
@@ -367,7 +387,9 @@ cpdef runSkipDataCaller(int masterIndex,  tcpMetadata,  table_obj,  vector[int]
     cdef vector[string] currentTableSchemaCppArgKeys
     cdef vector[string] currentTableSchemaCppArgValues
     cdef vector[string] tableNames
-    cdef vector[gdf_column_ptr] columns
+    cdef vector[column_view] columns
```
felipeblazing: this is a bit confusing to me why are there schemas, new_column and a table_view?

engine/src/io/data_parser/GDFParser.h
```diff
@@ -32,6 +33,7 @@ class gdf_parser : public data_parser {
 private:
 	std::vector<cudf::column *> columns;
```
felipeblazing: these lines shoudl be removed. we no longer keep a cudf::column * here

engine/src/execution_graph/logic_controllers/LogicPrimitives.h
```diff
@@ -58,6 +59,8 @@ class BlazingTableView {
 
 	std::vector<std::string> names() const;
 
+	void setNames(const std::vector<std::string> &names) { this->columnNames = names; }
```
felipeblazing: a view should have no way to set names. It is a view of a table. It is immutable.

engine/src/cython/io.cpp
```diff
@@ -71,9 +71,8 @@ TableSchema parseSchema(std::vector<std::string> files,
 		// TODO percy cudf0.12 port to cudf::column
 		//GDFRefCounter::getInstance()->deregister_column(column_cpp.get_gdf_column());
 		//tableSchema.columns.push_back(column_cpp.get_gdf_column());
-		
-		tableSchema.names.push_back(column_cpp.name());
 	}
+	tableSchema.blazingTableView.setNames(schema.get_names());
```
felipeblazing: views are immutable. I see why youa re doing this. You need to make a view and if theres no table behind it that doesnt really work. And we cant really create a table since there is no where to put it to keep it in scope.

engine/src/execution_graph/logic_controllers/LogicPrimitives.h
felipeblazing: I am guessing that this done because otherwise theres no way to manage the lifecycle.

engine/bsql_engine/io/cio.pxd
```diff
@@ -203,10 +203,7 @@ cdef extern from "../include/io/io.h":
         json_read_arg jsonReaderArg
         csv_read_arg csvReaderArg
     cdef struct TableSchema:
-        # TODO: TableSchema will be refactorized
-        # BlazingTableView blazingTableView
-        vector[gdf_column_ptr] columns
-        vector[string]  names
+        BlazingTableView blazingTableView
```
williamBlazing: If I am not mistaken we used to use a vector of gdf_column to contain the schema, where the columns would hold the types and it would also be the gdf data for a gdfTable. But I think now we want to have a vector of types and a vector of names to define the schema and have a separate BlazingTableView that is used ONLY for gdfTables. And the rest of the code should be updated to follow this new pattern

engine/bsql_engine/io/io.pyx
```diff
@@ -163,23 +163,31 @@ cpdef parseSchemaCaller(fileList, file_format_hint, args, extra_columns):
...
+
+    cdef int size = temp.blazingTableView.num_columns()
     i = 0
-    #for column in temp.columns:
-      # column.col_name = return_object['names'][i]
-      # return_object['columns'][return_object['names'][i].decode('utf-8')] = (gdf_column_to_column(column))
-      # i = i + 1
+    for column in range(0, size):
+      col_name = return_object['names'][i]
+      var = temp.blazingTableView.column(i)
```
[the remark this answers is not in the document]
Christian8491: Oh right, it's only in the parseSchemaCaller. Then is not necessary from here.

engine/bsql_engine/io/io.pyx
```diff
@@ -163,23 +163,31 @@ cpdef parseSchemaCaller(fileList, file_format_hint, args, extra_columns):
...
+      dtype = cudf_to_np_types[c.type().id()]
+      return_object['columns']['col_name'] = build_column(
```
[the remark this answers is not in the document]
Christian8491: but the second index level right? Ok just ->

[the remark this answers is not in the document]
felipeblazing: yes

engine/bsql_engine/io/io.pyx
```diff
@@ -163,23 +163,31 @@ cpdef parseSchemaCaller(fileList, file_format_hint, args, extra_columns):
...
+                cudf.core.Buffer(
+                    rmm.DeviceBuffer(ptr=<long long>var.data[void](), size=var.size() * dtype.itemsize),
```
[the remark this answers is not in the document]
Christian8491: Again is in the parseSchemaCaller, not necessary as above comment.

engine/bsql_engine/io/io.pyx
Christian8491: To can call the `view()` method in line 456, new_column should be a `Column` type. And you should explicitly said to cython the type.

engine/src/io/data_parser/GDFParser.h
Christian8491: Yep, I was waiting for the William PR for delete those attributes.

engine/bsql_engine/io/cio.pxd
Christian8491: Ohh. Ok that should change some lines of code ...

PR #270, opened 2020-01-22
[REVIEW] Changed evaluate_split_query, parse and load_data to return a TableViewPair

williamBlazing:


engine/src/execution_graph/logic_controllers/LogicPrimitives.h
```diff
@@ -57,6 +59,8 @@ class BlazingTableView {
 	cudf::column_view const & column(cudf::size_type column_index) const { return table.column(column_index); }
 
 	std::vector<std::string> names() const;
+	// set columnNames
+	void setNames(const std::vector<std::string> &names) { this->columnNames = names; }
```
felipeblazing: Views are immutable in terms of what columns they have you should never call set names. If you are this is an anti pattern

engine/src/io/data_parser/CSVParser.cpp
```diff
@@ -92,7 +91,8 @@ std::unique_ptr<ral::frame::BlazingTable> csv_parser::parse(
 	}
 
 	if(file == nullptr) { 
-		return create_empty_table(schema.get_names(), schema.get_dtypes(), column_indices);
+		// return create_empty_table(schema.get_names(), schema.get_dtypes(), column_indices);  // do we need to create an empty table that has metadata?
```
felipeblazing: file should never really be nullptr. Find out if anyone knows why it exists

engine/src/io/data_parser/CSVParser.cpp
```diff
@@ -115,22 +115,12 @@ std::unique_ptr<ral::frame::BlazingTable> csv_parser::parse(
...
-//				gdf_column_free(table_out.get_column(i));
-			} else {
-				//TODO create_gdf_column anymore
-				//columns_out[i].create_gdf_column(csv_table.tbl->get_column(i), csv_table.metadata.column_names[i]);
-			}
-		}
-		return std::make_unique<ral::frame::BlazingTable>(std::move(csv_table.tbl), csv_table.metadata.column_names);
+		std::unique_ptr<ral::frame::BlazingTable> table_out = std::make_unique<ral::frame::BlazingTable>(std::move(csv_table.tbl), csv_table.metadata.column_names);
+		ral::frame::BlazingTableView table_out_view = table_out->toBlazingTableView();
+		return std::make_pair(std::move(table_out), table_out_view);
```
felipeblazing: This is incorrect. If the unique_ptr to table is not nullptr then table_out_view is invalid.

felipeblazing: To be clear either the .first of the pair is nullptr or it is not, If it is not then we dont ever look at table_out_view. Doing it this way is VERY confusing. I am starting to think we should just make BlazingTable be able to be powered from both.

engine/src/io/data_parser/GDFParser.cpp
```diff
@@ -20,33 +20,34 @@ gdf_parser::gdf_parser(std::vector<cudf::column *> columns_, std::vector<std::st
 
 gdf_parser::~gdf_parser() {}
 
-std::unique_ptr<ral::frame::BlazingTable> gdf_parser::parse(
+ral::frame::TableViewPair gdf_parser::parse(
```
felipeblazing: No this cant be left off. This must be implemented here.

engine/src/io/data_parser/JSONParser.cpp
```diff
@@ -120,13 +120,13 @@ void json_parser::parse(std::shared_ptr<arrow::io::RandomAccessFile> file,
...
 
-std::unique_ptr<ral::frame::BlazingTable> json_parser::parse(
+ral::frame::TableViewPair json_parser::parse(
 	std::shared_ptr<arrow::io::RandomAccessFile> file,
 	const std::string & user_readable_file_handle,
 	const Schema & schema,
 	std::vector<size_t> column_indices) {
 		// TODO: Implement the new ARROW parser with 0.12 API whether already a new API exists for JSON parse
-		return nullptr;
+		return std::make_pair(nullptr, ral::frame::BlazingTableView());
```
felipeblazing: Why is the json_parser commented out?

engine/src/io/data_parser/OrcParser.cpp
```diff
@@ -75,22 +73,11 @@ std::unique_ptr<ral::frame::BlazingTable> orc_parser::parse(
...
-//				gdf_column_free(table_out.get_column(i));
-			} else {
-				//TODO create_gdf_column anymore
-//				columns_out[i].create_gdf_column(table_out.get_column(i));
-			}
-		}
-		return std::make_unique<ral::frame::BlazingTable>(std::move(orc_table.tbl), orc_table.metadata.column_names);
+		std::unique_ptr<ral::frame::BlazingTable> table_out = std::make_unique<ral::frame::BlazingTable>(std::move(orc_table.tbl), orc_table.metadata.column_names);
+		ral::frame::BlazingTableView table_out_view = table_out->toBlazingTableView();
+		return std::make_pair(std::move(table_out), table_out_view);
```
felipeblazing: either send the table or the view out not both.

engine/src/io/data_parser/ParquetParser.cpp
```diff
@@ -108,9 +105,11 @@ std::unique_ptr<ral::frame::BlazingTable> parquet_parser::parse(
 		//columns_out[i].create_gdf_column(table_out.get_column(i));
 		// }
 		// }
-		return std::make_unique<ral::frame::BlazingTable>(std::move(result.tbl), result.metadata.column_names);
+		std::unique_ptr<ral::frame::BlazingTable> table_out = std::make_unique<ral::frame::BlazingTable>(std::move(result.tbl), result.metadata.column_names);
+		ral::frame::BlazingTableView table_out_view = table_out->toBlazingTableView();
+		return std::make_pair(std::move(table_out), table_out_view);
```
felipeblazing: return the table or return it as nullptr with view set not both

engine/src/execution_graph/logic_controllers/LogicPrimitives.h
williamBlazing: When we do a load_data, we need to be able to set or change the names of the columns after we have loaded them. This function was added in order to be able to do that. There is no way to change the names of a BlazingTable. The alternative is to pass in aliases into load_data, which would mean we would also have to add it to parse, which would mean that this alias handling code would have to be added to 14 files

williamBlazing: I agree that BlazingTable should be immutable, but I really dont like the idea of having to deal with so much to make it work. Also, its the data that is immutable, i dont think we need to be so strict with the names. I dont think that function would get misused

engine/src/io/data_parser/CSVParser.cpp
williamBlazing: Yeah, I dont know why either, we can easily take it out

williamBlazing: That was done over 5 month ago, i doubt if anyone will remember

engine/src/io/data_parser/CSVParser.cpp
williamBlazing: Usually what we actually use are BlazingTableViews in the code, BlazingTable is just an owner object. So the function that receives the pair will want to use the BlazingTableView more than the BlazingTable. So if the pattern is that you either use one OR the other, then on the receiving end you would always need to do something like this:
```TableViewPair pair = function()
BlazingTableView my_view = pair.first == nullptr ? pair.second : pair.first.toBlazingTableVIew();```

So we either want every receiving function to handle that, or every function that outputs a pair to make the BlazingTableView for you. I think its better for the function that outputs a pair to handle that.

Also, you dont want the receiving function to be using BLazingTable and the main object it uses, because you dont want to be null checking BlazingTable every single time you want to use it.

engine/src/io/data_parser/GDFParser.cpp
williamBlazing: This will be implemented once @Christian8491 's PR is complete which has the updated Schema object that contains a BlazingTableView

williamBlazing: https://github.com/BlazingDB/blazingsql/pull/268

engine/src/io/data_parser/JSONParser.cpp
williamBlazing: I believe it has not been migrated yet

engine/src/io/data_parser/ParquetParser.cpp
williamBlazing: see comments above

engine/src/execution_graph/logic_controllers/LogicPrimitives.h
felipeblazing: im ok with it for now but I think this will need revisiting

engine/src/io/data_parser/CSVParser.cpp
felipeblazing: Yes I am advocating for doing exactly what you describe instead

PR #276, opened 2020-01-23
[REVIEW] Add support for string comparison in interops

### LINKED ISSUE #278: Migration string comparison in interops

jeanp413:
Closes #278


engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -152,512 +114,264 @@ public:
...
-	template <typename ColType, typename LocalStorageType>
-	__device__ __forceinline__ void device_ptr_read_into_buffer(column_index_type col_index,
-		cudf::size_type row,
-		const void * const * columns,
-		int64_t * buffer,
-		int position) {
-		const ColType * col_data = static_cast<const ColType *>((columns[col_index]));
-		*((LocalStorageType *) (buffer + ((position * threadBlockSize) + threadIdx.x))) =
-			(LocalStorageType) __ldg(((ColType *) &col_data[row]));
+	struct device_ptr_read_into_buffer {
```
felipeblazing: I prefer the template where there is a typename for the LocalStorageType it makes it much clearer what is happening

engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -677,253 +391,358 @@ private:
...
-
-			gdf_unary_operator oper = this->unary_operations[op_index];
+			get_data_from_buffer(&left_value, buffer, left_position);
+			bool left_valid = getColumnValid(row_valids, left_position);
+			
+			gdf_unary_operator oper = unary_operations[op_index];
 			if (left_valid) {
 				if(oper == BLZ_FLOOR) {
-					computed = floor(left_value);
+					double val = static_cast<double>(left_value);
```
felipeblazing: why are you making assumptions about types that we were not making before. What was wrong with LeftType left_value and RightType and OutputType?

engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -677,253 +391,358 @@ private:
...
+					store_data_in_buffer(floor(val), buffer, output_position);
 				} else if(oper == BLZ_CEIL) {
-					computed = ceil(left_value);
+					double val = static_cast<double>(left_value);
```
felipeblazing: the previous code seems easier to follow most operations are just 
```
computed = operation(left_value);
```

engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -677,253 +391,358 @@ private:
...
-	int64_t * scalars_right;
-
-	cudf::size_type * null_counts_inputs;
-	cudf::size_type * null_counts_outputs;
-
-	int64_t * temp_valids_in_buffer;
-	int64_t * temp_valids_out_buffer;
-
-	int threadBlockSize;
+	cudf::mutable_table_device_view out_table;
```
felipeblazing: How are you ensuring that this information is available to the gpu?

felipeblazing: How are you making sure that this is all in one contiguous buffer?

engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -152,512 +114,264 @@ public:
...
-			if(right_index == UNARY_INDEX) {
-				// unary
-				type_from_op = get_output_type(left_input_types_vec[i], unary_operators[i]);
-			} else {
-				// binary
-				type_from_op = get_output_type(left_input_types_vec[i], right_input_types_vec[i], operators[i]);
-			}
+		void * temp_valids_out_buffer)
+		: out_table{out_table},
+			table{table},
```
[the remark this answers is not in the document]
jeanp413: Not anymore since tables and scalars are passed in a different way I ended up deleting most of the data in `temp_space`  buffer. For now I created `rmm::device_vector`s for the remaining data and I pass the pointer to the functor, when the migration is completed we could run some benchmarks and go back to one unique buffer for the remaining data if that's the case

engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -152,512 +114,264 @@ public:
...
+		cudf::bitmask_type * valids_in_buffer =
+			temp_valids_in_buffer + (blockIdx.x * blockDim.x + threadIdx.x) * table.num_columns();
+		cudf::bitmask_type * valids_out_buffer =
+			temp_valids_out_buffer + (blockIdx.x * blockDim.x + threadIdx.x) * out_table.num_columns();
 
-		int64_t cur_row_valids;
-		for(cudf::size_type row = 0; row < 64 && row_index + row < size; row++) {
-			load_cur_row_valids(valids_in_buffer, row, cur_row_valids, this->num_columns);
+		for(cudf::size_type column_index = 0; column_index < table.num_columns(); column_index++) {
+			read_valid_data(column_index, valids_in_buffer, row_index);
```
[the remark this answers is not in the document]
jeanp413: No `load_cur_row_valids` is the `setColumnValid` [here](https://github.com/BlazingDB/blazingsql/pull/276/files?file-filters%5B%5D=.cuh#diff-ce59fe5a52cbeb027fd56f962a93b7e4R167) it's just inlined

engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -152,512 +114,264 @@ public:
...
+			for(int16_t op_index = 0; op_index < num_operations; op_index++) {
 				process_operator(op_index, total_buffer, row_index + row, cur_row_valids);
 			}
 
-			for(int out_index = 0; out_index < this->num_final_outputs; out_index++) {
-				write_data(out_index, this->final_output_positions[out_index], total_buffer, row_index + row);
+			// copy data and row valids into buffer
+			for(cudf::size_type column_index = 0; column_index < out_table.num_columns(); column_index++) {
+				write_data(column_index, final_output_positions[column_index], total_buffer, row_index + row);
+				setColumnValid(valids_out_buffer[column_index],	row, getColumnValid(cur_row_valids, this->final_output_positions[column_index]));
```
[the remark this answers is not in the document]
jeanp413: It's the same just inlined the functions calls

engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -152,512 +114,264 @@ public:
...
+
+	CUDA_DEVICE_CALLABLE void get_data_from_buffer(double * data, int64_t * buffer, int position) {
+		int64_t * register_ptr = buffer + (position * blockDim.x + threadIdx.x);
+		*data = __longlong_as_double(*register_ptr);
 	}
 
-	template <typename LocalStorageType>
-	__device__ __forceinline__ void store_data_in_buffer(LocalStorageType data, int64_t * buffer, int position) {
-		*((LocalStorageType *) (buffer + ((position * threadBlockSize) + threadIdx.x))) = data;
+	CUDA_DEVICE_CALLABLE void store_data_in_buffer(int64_t data, int64_t * buffer, int position) {
```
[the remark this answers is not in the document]
jeanp413: Now I'm using `__double_as_longlong` when writing a `double` to a `int64_t`, I could use template specialization but thought this was simpler

engine/src/Interpreter/interpreter_ops.cuh
jeanp413: I need to specialize for the different data types, the buffer is always `int64_t *` so no need for another tempate argument, just need to apply the correct conversion like `__double_as_longlong`

engine/src/Interpreter/interpreter_ops.cuh
jeanp413: OutputType can be inferred from the inputs no need for another template argument in the case of `floor` it only applies to `double` if you pass an `int64_t` it would be converted to double

engine/src/Interpreter/interpreter_ops.cuh
jeanp413: same  as above just making it explicit

engine/src/Interpreter/interpreter_ops.cuh
jeanp413: Not sure what you mean,  this class is meant to be used in kernels as it's name implies

engine/src/Interpreter/interpreter_ops.cuh
felipeblazing: There are many reasons why someone controlling the code would want to say exactly what the output type should be. I dont think it should be inferred.

engine/src/Interpreter/interpreter_ops.cuh
felipeblazing: this is more lines of code and both do the same thing.

engine/src/Interpreter/interpreter_ops.cuh
felipeblazing: ok i guess we arent making sure of that anymore. We need to do some performance testing to see if there is a cost to this. Because the plan was all in one contiguous buffer before you knew that every thread was basically reading from cache when looking at the plan.

engine/src/Interpreter/interpreter_ops.cuh
jeanp413: The final output is where the conversion to different types is done. All the intermediate results are
either int64_t or double, it's the same as before.

engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -152,512 +114,264 @@ public:
...
+private:
 
 	/**
 	 * @param buffer the local buffer which storse the information that is to be processed
 	 * @param position the position in the local buffer where this data needs to be written
 	 */
-	template <typename LocalStorageType>
-	__device__ __forceinline__ LocalStorageType get_data_from_buffer(int64_t * buffer, int position) {
-		return *((LocalStorageType *) (buffer + ((position * threadBlockSize) + threadIdx.x)));
+	CUDA_DEVICE_CALLABLE void get_data_from_buffer(int64_t * data, int64_t * buffer, int position) {
```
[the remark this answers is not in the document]
jeanp413: done

PR #332, opened 2020-01-31
[REVIEW] Cudf 0.12 migration with gpuci integration

mario21ic:


aucahuasi:
@mario21ic delete */conda (algebra/conda, engine/conda ...etc) folders please since we should not use anymore those old scripts

PR #341, opened 2020-02-03
[REVIEW] adding a variable to upload into anaconda cloud

mario21ic:


ci/cpu/upload_anaconda.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+SOURCE_BRANCH="branch-0.12"
```
dillon-cullinan: This needs to be `master`

PR #355, opened 2020-02-04
[REVIEW] Fix samples by ratio, fix group by with aggregations, fix interops output type

williamBlazing:


engine/src/cuDF/generator/sample_generator.cu
```diff
@@ -72,10 +72,10 @@ std::unique_ptr<ral::frame::BlazingTable> generate_sample(
 
 	cudf::generator::RandomVectorGenerator<std::int32_t> generator(0L, num_rows);
 	std::vector<std::int32_t> arrayIdx = generator(num_samples);
-
+	
 	rmm::device_buffer gatherData(arrayIdx.data(), num_samples * sizeof(std::int32_t));
 
-	cudf::column gatherMap{cudf::data_type{cudf::type_id::INT32}, num_samples, gatherData};
+	cudf::column gatherMap{cudf::data_type{cudf::type_id::INT32}, num_samples, std::move(gatherData)};
```
felipeblazing: what will happen to this code when the indices change from and int32 to an int64. Is there a way of doing this where we tie it to whatever cudf::sizetype is? like whatever we use for indexing the rows.

engine/src/execution_graph/logic_controllers/LogicalProject.cpp
```diff
@@ -17,6 +17,7 @@
 #include "../../Interpreter/interpreter_cpp.h"
 #include "../../parser/expression_tree.hpp"
 #include "../../Utils.cuh"
+#include "utilities/DebuggingUtils.h"
```
felipeblazing: this is probable left over from debugging im guessing :)

engine/src/operators/GroupBy.cpp
```diff
@@ -9,6 +9,7 @@
 #include "distribution/primitives.h"
 #include "utilities/CommonOperations.h"
 #include "utilities/RalColumn.h"
+#include "utilities/DebuggingUtils.h"
```
felipeblazing: im guessing this is not meant to be in here either

engine/src/execution_graph/logic_controllers/LogicalProject.cpp
aucahuasi: yep removed thanks

engine/src/operators/GroupBy.cpp
aucahuasi: removed thanks

engine/src/cuDF/generator/sample_generator.cu
aucahuasi: I have some changes here but I think the best approach for now is to leave this as is since handling arbitrary size could changes another functions too

aucahuasi: cc @felipeblazing @williamBlazing I will add this to my TODOs list

PR #371, opened 2020-02-06
[REVIEW] Implement case for strings

jeanp413:


engine/tests/new-tests/process_project.cu
```diff
@@ -301,3 +301,21 @@ TEST_F(ProjectTestString, test_cast_from_string)
 
     cudf::test::expect_tables_equal(expected_table_view, out_table->view());
 }
+
+TEST_F(ProjectTestString, test_string_case)
```
felipeblazing: can you add a few more tests that make sure that this case still works when its inside of other expressions?

aucahuasi: was done, thanks!

aucahuasi:
@felipeblazing Case(...) + Case(...) was ok @jeanp413 add that in a unit test, thanks

PR #374, opened 2020-02-07
[REVIEW] added selection of columns when returning empty tables from data_loader

williamBlazing:


engine/src/distribution/primitives.cpp
```diff
@@ -219,14 +219,6 @@ std::vector<NodeColumnView> partitionData(Context * context,
 
 	std::vector<CudfTableView> partitioned_data = cudf::experimental::split(table.view(), host_pivot_indexes.first);
 
-	std::unique_ptr<CudfTable> empty;
```
felipeblazing: Did you verify that this is no longer an issue?

PR #390, opened 2020-02-11
[TESTING] finishing migrating hive

williamBlazing:
This should implement Hive functionality, but is still missing the full skip-data stuff.

Its worth noting a couple mayor changes. 
We used to send vectors of scalars, strings and bools to represent the partition values from hive. The scalars were the values, the bools would indicate if the scalars were strings or not, and if they were they would be captured in a string. This was really messy and fortunately we could not migrate that mess anyways since there are no new cudf::scalars on the python side for libcudf++ anyways. So I simplified it all by having it all as strings and converting those strings to the appropriate scalar-filled columns on the C++ side anyways.


williamBlazing:
this work was continued on a different PR

PR #442, opened 2020-03-03
[REVIEW] build: remove cython cache files

mario21ic:


build.sh
```diff
@@ -124,11 +124,11 @@ if buildAll || hasArg io; then
           -DBUILD_TESTING=${TESTS} \
           -DCMAKE_BUILD_TYPE=${BUILD_TYPE} ..
 
+    IO_BLD=""
```
aucahuasi: Why we need this var?

build.sh
```diff
@@ -143,11 +143,11 @@ if buildAll || hasArg comms; then
           -DBUILD_TESTING=${TESTS} \
           -DCMAKE_BUILD_TYPE=${BUILD_TYPE} ..
```
aucahuasi: where is the check if case of errors?

PR #532, opened 2020-04-09
[TESTING] limiting number of file handles opened

williamBlazing:

PR #538, opened 2020-04-10
[TESTING] create_table with user defined partitions

### LINKED ISSUE #534: Add partition specification to create_table
Proposal:

bc.create_table(‘table name’, cursor, schema_name, table_name, partitions= {}, filetype);

The partitions parameter contains the partition name and value that we are looking for in the key, value pairs.
 

Example: we have partition on `col1` and `col1` has values from 1 to 1000, and we are just interested in values 10 to 19.

Our create table would look something like this,

bc.create_table(‘table name’, cursor, schema_name, table_name, partitions= {col1:[I for I in range(10:20)] }, filetype);

 doing this we do not have scan the filenames for rest of the folders/partitions.

Taking this a step further, we can skip the whole concept of pyhive cursor by just doing a create table with `bc`, keberos auth using the below prototype.

bc.create_table(‘table name’, path=’path to baser folder dir’, partitions= {}, filetype);
 

Here the partitions parameter contains partition names and values as key, value pairs. But in this case we have to make sure they follow the exact same order and all the partitions are present in keys.
 

Example:

We have partitions on `col1` and `col1` has values from 1 to 1000, and we are just interested in values 10 to 19.

Also the table is partitioned on `col2` and we need all the values. And the partitions of hive are present in the same order.

 

Our create table would look something like this,

bc.create_table(‘table name’, path=’path to baser folder dir’, dict= {col1:[I for I in range(10:20)], col2:[]}, filetype);

williamBlazing:
This resolves #534

### DIFF
--- a/engine/bsql_engine/io/cio.pxd
+++ b/engine/bsql_engine/io/cio.pxd
@@ -109,7 +109,7 @@ cdef extern from "../include/io/io.h":
     pair[bool, string] registerFileSystemGCS( GCS gcs, string root, string authority) except +raiseRegisterFileSystemGCSError
     pair[bool, string] registerFileSystemS3( S3 s3, string root, string authority) except +raiseRegisterFileSystemS3Error
     pair[bool, string] registerFileSystemLocal(  string root, string authority) except +raiseRegisterFileSystemLocalError
-    TableSchema parseSchema(vector[string] files, string file_format_hint, vector[string] arg_keys, vector[string] arg_values, vector[pair[string,type_id]] types) except +raiseParseSchemaError
+    TableSchema parseSchema(vector[string] files, string file_format_hint, vector[string] arg_keys, vector[string] arg_values, vector[pair[string,type_id]] types, bool ignore_missing_paths) except +raiseParseSchemaError
     unique_ptr[ResultSet] parseMetadata(vector[string] files, pair[int,int] offsets, TableSchema schema, string file_format_hint, vector[string] arg_keys, vector[string] arg_values) except +raiseParseSchemaError
 
 cdef extern from "../src/execution_graph/logic_controllers/LogicPrimitives.h" namespace "ral::frame":
--- a/engine/src/io/DataLoader.cpp
+++ b/engine/src/io/DataLoader.cpp
@@ -213,20 +213,24 @@ std::unique_ptr<ral::frame::BlazingTable> data_loader::load_data(
 
 
 void data_loader::get_schema(Schema & schema, std::vector<std::pair<std::string, cudf::type_id>> non_file_columns) {
-	std::vector<std::shared_ptr<arrow::io::RandomAccessFile>> files;
-	std::vector<data_handle> handles = this->provider->get_some(5); // we only need one to get the schema, but we will get a couple in-case the first couple files are empty
-	for(auto handle : handles) {
-		files.push_back(handle.fileHandle);
+	bool got_schema = false;
+	while (!got_schema && this->provider->has_next()){
+		data_handle handle = this->provider->get_next();
+		if (handle.fileHandle != nullptr){
+			this->parser->parse_schema(handle.fileHandle, schema);
+			if (schema.get_num_columns() > 0){
+				got_schema = true;
+				schema.add_file(handle.uri.toString(true));
+			}
+		}		
 	}
-	this->parser->parse_schema(files, schema);
-
-	for(auto handle : handles) {
-		schema.add_file(handle.uri.toString(true));
+	if (!got_schema){
+		std::cout<<"ERROR: Could not get schema"<<std::endl;
 	}
-
+	
 	bool open_file = false;
 	while (this->provider->has_next()){
-		handles = this->provider->get_some(32, open_file);
+		std::vector<data_handle> handles = this->provider->get_some(64, open_file);
 		for(auto handle : handles) {
 			schema.add_file(handle.uri.toString(true));
 		}		
@@ -240,7 +244,7 @@ void data_loader::get_schema(Schema & schema, std::vector<std::pair<std::string,
 
 std::unique_ptr<ral::frame::BlazingTable> data_loader::get_metadata(int offset) {
 	
-	std::size_t NUM_FILES_AT_A_TIME = 32;
+	std::size_t NUM_FILES_AT_A_TIME = 64;
 	std::vector<std::unique_ptr<ral::frame::BlazingTable>> metadata_batches;
 	std::vector<ral::frame::BlazingTableView> metadata_batche_views;
 	while(this->provider->has_next()){
--- a/engine/src/io/DataLoader.cpp
+++ b/engine/src/io/DataLoader.cpp
@@ -213,6 +213,7 @@ std::unique_ptr<ral::frame::BlazingTable> data_loader::load_data(
 
 
 void data_loader::get_schema(Schema & schema, std::vector<std::pair<std::string, cudf::type_id>> non_file_columns) {
+	std::cout<<"get_schema start"<<std::endl;
 	bool got_schema = false;
 	while (!got_schema && this->provider->has_next()){
 		data_handle handle = this->provider->get_next();
@@ -227,19 +228,22 @@ void data_loader::get_schema(Schema & schema, std::vector<std::pair<std::string,
 	if (!got_schema){
 		std::cout<<"ERROR: Could not get schema"<<std::endl;
 	}
+	std::cout<<"get_schema got schema"<<std::endl;
 	
 	bool open_file = false;
 	while (this->provider->has_next()){
 		std::vector<data_handle> handles = this->provider->get_some(64, open_file);
 		for(auto handle : handles) {
 			schema.add_file(handle.uri.toString(true));
-		}		
+		}
+		std::cout<<"get_schema got some file paths"<<std::endl;
 	}
 
 	for(auto extra_column : non_file_columns) {
 		schema.add_column(extra_column.first, extra_column.second, 0, false);
 	}
 	this->provider->reset();
+	std::cout<<"get_schema end"<<std::endl;
 }
 
 std::unique_ptr<ral::frame::BlazingTable> data_loader::get_metadata(int offset) {
--- a/engine/src/io/DataLoader.cpp
+++ b/engine/src/io/DataLoader.cpp
@@ -48,15 +48,14 @@ std::unique_ptr<ral::frame::BlazingTable> data_loader::load_data(
 		std::iota(column_indices.begin(), column_indices.end(), 0);
 	}
 
-	std::vector<std::string> user_readable_file_handles;
 	std::vector<data_handle> files;
 
 	// iterates through files and parses them into columns
 	this->provider->reset();
 	while(this->provider->has_next()) {
-		// a file handle that we can use in case errors occur to tell the user which file had parsing issues
-		user_readable_file_handles.push_back(this->provider->get_current_user_readable_file_handle());
-		files.push_back(this->provider->get_next());
+		auto handle = this->provider->get_next();
+		if (handle.is_valid())
+			files.emplace_back(std::move(handle));
 	}
 	size_t num_files = files.size();
 
@@ -94,7 +93,7 @@ std::unique_ptr<ral::frame::BlazingTable> data_loader::load_data(
 					}
 
 					if (schema.all_in_file()){
-						std::unique_ptr<ral::frame::BlazingTable> loaded_table = parser->parse(file_sets[file_set_index][file_in_set].fileHandle, user_readable_file_handles[file_index], fileSchema, column_indices_in_file);
+						std::unique_ptr<ral::frame::BlazingTable> loaded_table = parser->parse(file_sets[file_set_index][file_in_set].fileHandle, fileSchema, column_indices_in_file);
 
 						if(filterString != ""){
 							blazingTable_per_file[file_index] = std::move(ral::processor::process_filter(loaded_table->toBlazingTableView(), filterString, context));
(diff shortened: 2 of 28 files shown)


engine/bsql_engine/io/cio.pxd
```diff
@@ -109,7 +109,7 @@ cdef extern from "../include/io/io.h":
     pair[bool, string] registerFileSystemGCS( GCS gcs, string root, string authority) except +raiseRegisterFileSystemGCSError
     pair[bool, string] registerFileSystemS3( S3 s3, string root, string authority) except +raiseRegisterFileSystemS3Error
     pair[bool, string] registerFileSystemLocal(  string root, string authority) except +raiseRegisterFileSystemLocalError
-    TableSchema parseSchema(vector[string] files, string file_format_hint, vector[string] arg_keys, vector[string] arg_values, vector[pair[string,type_id]] types) except +raiseParseSchemaError
+    TableSchema parseSchema(vector[string] files, string file_format_hint, vector[string] arg_keys, vector[string] arg_values, vector[pair[string,type_id]] types, bool ignore_missing_paths) except +raiseParseSchemaError
```
felipeblazing, line 112: why is this a configuration. Who sets this ignore_missing_paths?

engine/src/io/DataLoader.cpp
```diff
@@ -94,7 +93,7 @@ std::unique_ptr<ral::frame::BlazingTable> data_loader::load_data(
 					}
 
 					if (schema.all_in_file()){
-						std::unique_ptr<ral::frame::BlazingTable> loaded_table = parser->parse(file_sets[file_set_index][file_in_set].fileHandle, user_readable_file_handles[file_index], fileSchema, column_indices_in_file);
+						std::unique_ptr<ral::frame::BlazingTable> loaded_table = parser->parse(file_sets[file_set_index][file_in_set].fileHandle, fileSchema, column_indices_in_file);
```
felipeblazing, line 96: why did you get rid of user_readable_file_handles? the point of this was to be able to let us tell the user on WHICH file they had an issue and make that file path readable to the user.

engine/src/io/DataLoader.cpp
```diff
@@ -149,7 +148,7 @@ std::unique_ptr<ral::frame::BlazingTable> data_loader::load_data(
 					}
 				} else {
 					Library::Logging::Logger().logError(ral::utilities::buildLogString(
-						"", "", "", "ERROR: Was unable to open " + user_readable_file_handles[file_index]));
+						"", "", "", "ERROR: Was unable to open " + file_sets[file_set_index][file_in_set].uri.toString()));
```
felipeblazing, line 151: There was a reason to keep this seperate before. Because the handle we have is NOT the same as what the user registered it as on their side when they made the filesystem

engine/src/io/DataLoader.cpp
```diff
@@ -215,8 +214,7 @@ std::unique_ptr<ral::frame::BlazingTable> data_loader::load_data(
 
 void data_loader::get_schema(Schema & schema, std::vector<std::pair<std::string, cudf::type_id>> non_file_columns) {
 	std::vector<std::shared_ptr<arrow::io::RandomAccessFile>> files;
-	bool firstIteration = true;
-	std::vector<data_handle> handles = this->provider->get_all();
+	std::vector<data_handle> handles = this->provider->get_some(5); // we only need one to get the schema, but we will get a couple in-case the first couple files are empty
```
felipeblazing, line 217: This is really not correct. Don't get some and assume its fine. get one by one until one is not empty.

engine/src/io/DataLoader.cpp
```diff
@@ -226,23 +224,42 @@ void data_loader::get_schema(Schema & schema, std::vector<std::pair<std::string,
 		schema.add_file(handle.uri.toString(true));
 	}
 
+	bool open_file = false;
+	while (this->provider->has_next()){
+		handles = this->provider->get_some(32, open_file);
```
felipeblazing, line 229: what is the logic behind getting 32 at a time?

engine/src/io/DataLoader.cpp
```diff
@@ -226,23 +224,42 @@ void data_loader::get_schema(Schema & schema, std::vector<std::pair<std::string,
...
 std::unique_ptr<ral::frame::BlazingTable> data_loader::get_metadata(int offset) {
-	std::vector<std::shared_ptr<arrow::io::RandomAccessFile>> files;
-
-	bool firstIteration = true;
-	std::vector<data_handle> handles = this->provider->get_all();
-
-	for(auto handle : handles) {
-		files.push_back(handle.fileHandle);
+	
+	std::size_t NUM_FILES_AT_A_TIME = 32;
```
felipeblazing, line 243: Why is this number set to 32? How did you arrive at this "optimal" number

engine/src/io/DataLoader.cpp
```diff
@@ -226,23 +224,42 @@ void data_loader::get_schema(Schema & schema, std::vector<std::pair<std::string,
...
+	std::vector<std::unique_ptr<ral::frame::BlazingTable>> metadata_batches;
+	std::vector<ral::frame::BlazingTableView> metadata_batche_views;
+	while(this->provider->has_next()){
+		std::vector<std::shared_ptr<arrow::io::RandomAccessFile>> files;
+		std::vector<data_handle> handles = this->provider->get_some(NUM_FILES_AT_A_TIME);
+		for(auto handle : handles) {
+			files.push_back(handle.fileHandle);
+		}
+		metadata_batches.emplace_back(this->parser->get_metadata(files,  offset));
```
felipeblazing, line 252: what is this metadata logic that is being introduced here?

engine/src/io/data_provider/DummyProvider.h
```diff
@@ -34,27 +34,27 @@ class dummy_data_provider : public data_provider {
 		// does nothing
 	}
 
-	data_handle get_next() {
+	data_handle get_next(bool open_file = true) {
```
felipeblazing, line 37: I am really starting to dislike this logic where you tell it open th efile to set some state to true or not true. In reality what wewant here is to reset the counter after we have run get_next once of a few times. The proper way to do this is to make an api that is .reset_provider or somethign like this. It seems to be poor form to have get_next operate these two ways. Another way would be to do something like peek() instead of get_next if you wanted to get some of the data but not change the internal state. This just seems a bit confusing to me as a user.

if I saw

handle.get_next(true), handle.get_next(false) I woul dhave to go into the function signature to see what it was trying to do. If instead you did

```
handle.get_next() -- get me the next one and change state
handle.peek(3)  -- get me the next 3 and change no state
```

this becomes much clearer

engine/src/io/data_provider/DummyProvider.h
```diff
@@ -34,27 +34,27 @@ class dummy_data_provider : public data_provider {
...
 
-	std::string get_current_user_readable_file_handle() { return ""; }
-
-	std::vector<data_handle> get_all() { return {}; }
-
+	/**
+	 * Tries to get up to num_files data_handles. We use this instead of a get_all() because if there are too many files, 
+	 * trying to get too many file handles will cause a crash. Using get_some() forces breaking up the process of getting file_handles.
+	 */
+	std::vector<data_handle> get_some(std::size_t num_files, bool open_file = true) { return {}; }
```
felipeblazing, line 49: When you made this change you need to ask yourself HOW am I deciding how many files at at time. What was caussing the crashes  you were trying to resolve? Were you running out of file handles? Maybe you just need to reuse them? This seems to be a way to stop the crashes right now but it doesn't tell us if we solved the underlying issue. Also this needs to be moved into its own PR it is a bug and not related to this functionality.

engine/src/io/data_provider/UriDataProvider.cpp
```diff
@@ -17,16 +17,17 @@
 namespace ral {
 namespace io {
 
-uri_data_provider::uri_data_provider(std::vector<Uri> uris)
+uri_data_provider::uri_data_provider(std::vector<Uri> uris, bool ignore_missing_paths)
```
felipeblazing, line 20: what does it mean to ignore missing paths? Does it mean if a path does not exist but the user said it did we will pretends as if it had? This doesnt make much sense to me.

engine/src/io/data_provider/UriDataProvider.cpp
```diff
@@ -117,7 +114,7 @@ data_handle uri_data_provider::get_next() {
 
 			if(fs_manager && fs_manager->exists(target_uri)) {
 				fileStatus = BlazingContext::getInstance()->getFileSystemManager()->getFileStatus(target_uri);
-			} else {
+			} else if (!ignore_missing_paths){
```
felipeblazing, line 117: why would we not want to tell the user we couldnt find a file they expected?

engine/src/io/data_provider/UriDataProvider.cpp
```diff
@@ -186,17 +185,25 @@ data_handle uri_data_provider::get_next() {
 			return handle;
 		} else {
 			// this is a file we cannot parse apparently
-			return get_next();
+			this->current_file++;
```
felipeblazing, line 188: we shoudl be throwing some kind of warning here that we can somehow let the user know that something was aq little off. This might be one of the cases where a cout warning like the red ones that dask sometimes gives users would be the most appropriate

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -1158,7 +1200,7 @@ def drop_table(self, table_name):
         """
         self.add_remove_table(table_name, False)
 
-    def _parseSchema(self, input, file_format_hint, kwargs, extra_columns):
+    def _parseSchema(self, input, file_format_hint, kwargs, extra_columns, ignore_missing_paths):
```
felipeblazing, line 1203: when and how is ignore_missing_pathjs set

engine/bsql_engine/io/cio.pxd
williamBlazing: The ignore_missing_paths feature was added because we are wanting to allow a user to specify  a set of partitions for a table. This means we need to constructed the expected paths, but not all partition combinations may exist, which means we need to be ok with contructing a path that may not actually exist, and therefore not really be an error. For example:
Say the user requests:  partitions:{'colA' :[1, 2], 'colB':[3, 4]}
So from this we build four directories:
base_folder/colA=1/colB=3/*
base_folder/colA=1/colB=4/*
base_folder/colA=2/colB=3/*
base_folder/colA=2/colB=4/*
But if the partition combination colA=1/colB=4 did not actually exist, then that path does not exist and its ok. We dont want to throw an error. In order to enable that I introduced that ignore_missing_paths functionality. It only ever gets used during `parseSchema` and only if the user is specifying partitions without using Hive.

engine/src/io/DataLoader.cpp
williamBlazing: user_readable_file_handles was not being used in any of the code except in one place to show an error. When we introduced the `data_handles` struct which contains the Uri, then user_readable_file_handles became useless and should have been taken out

engine/src/io/DataLoader.cpp
williamBlazing: user_readable_file_handles[file_index] in this context and file_sets[file_set_index][file_in_set].uri.toString() are created the exact same way, I am not sure what you are talking about. You may be thinking of how we would like for this to work, not of how it was actually working

engine/src/io/DataLoader.cpp
williamBlazing: Good point, i will fix this

engine/src/io/DataLoader.cpp
williamBlazing: Honestly, I just picked a number. Any suggestions?

engine/src/io/data_provider/DummyProvider.h
williamBlazing: You are correct, if should be in its own PR and it is https://github.com/BlazingDB/blazingsql/pull/532
This PR has both because I wanted the user that requested this to test out both things.
That same user will tell us if this will indeed fix the issue we were seeing.
Either way, we do know that allowing the engine to open an unlimited number of file handles at a time will cause problems, so I think that we do want to implement a way to limit the number of file handles that are opened at a time and this is one way to deal with that.
Now as for deciding how many to open at a time, I dont not have a good answer to that. I am open to suggestions

engine/src/io/data_provider/UriDataProvider.cpp
williamBlazing: See my comment above

engine/src/io/data_provider/UriDataProvider.cpp
williamBlazing: See my comment above

engine/src/io/data_provider/UriDataProvider.cpp
williamBlazing: The code only gets here if we are purposefully ignoring missing files. See my explanation above for why we are doing that. The code in this if/else branch is more about permitting continuing the traversal of the files and directories by the provider

pyblazing/pyblazing/apiv2/context.py
williamBlazing: See my comment above

engine/bsql_engine/io/cio.pxd
felipeblazing: ok I see so the actual folders coudl be missing and it would throw an error. Yah I cant think of another way around this. What about the case when the user fucks up and NONE of the path exist? Or like I specify

col a = [part1] col b [part2] and there are NO values for anyone in part1 so it would be empty, do we want to let them konw that?, or even worse what if the user specifies col abc when no column abc exists?

engine/src/io/DataLoader.cpp
felipeblazing: ok

engine/src/io/DataLoader.cpp
felipeblazing: I figure the number should be as high as can be without triggering issues. So to figure this out we need to see when the issues start occurring in different contexts. So to b eclear, the worst case is when 1 machine has the most amount of gpus. so with 8 or 16 you are generating 8 or 16 x the number of file handles. Set up something like 16 dask nodes on one machine then see when it starts crapping out. Pick some percentage of this as the top. I would say something like 10% of when it craps out. Or if I am going to rnadomly guess I would go for osmething like 10k 

This is of course in the case that its reading locally. If we are using  another file system like hdfs we need to see what error was happening.

engine/bsql_engine/io/cio.pxd
williamBlazing: Ok, so if the user screwed up and like col abc does not exist, then it will find no folders and no files when trying to get the schema, in which case it should throw an error there. I will test out that scenario and make sure it works

engine/src/io/DataLoader.cpp
williamBlazing: So, a starting point the default max number of file handles you can have open is 1024. So 1024/16 = 64.
In hdfs the default is 128 per process
So, given some preliminary research i can set it to 64 instead of 32

engine/src/io/DataLoader.cpp
williamBlazing: The metadata collection takes a set of file handles and a `offset` (what is the starting file index) and gives you a metadata dataframe. Now that we can only have a set of files only at a time, we need to then collect the metadata in batches and concatenate the metadata

PR #571, opened 2020-04-21
Update README.md

gumdropsteve:
- inline link instead of colab button
- add "Try Now" to to menu
- 2 -> two


GPUtester:

Issue #566, opened 2020-04-20
[BUG] Dask Delayed with Blazing causing Exceptions


mlahir1:
Dask Delayed with Blazing causing Exceptions

### Case 1
Code:
```
import cudf, glob
from blazingsql import BlazingContext
from dask.distributed import Client
from dask_cuda import LocalCUDACluster
from dask.distributed import Client
from dask.delayed import delayed
import dask

cluster = LocalCUDACluster()
client = Client(cluster)

def fun():
    bc = BlazingContext()
    bc.create_table('tab', cudf.DataFrame({'a':range(10000)}))
    return bc.sql('select * from tab')
    
x = [delayed(fun)() for i in range(10)]
dask.compute(x)
```
Exception:
```
<ipython-input-2-674e6ae578f9> in fun()
     13     bc = BlazingContext()
     14     bc.create_table('tab', cudf.DataFrame({'a':range(10000)}))
---> 15     return bc.sql('select * from tab')
     16 
     17 x = [delayed(fun)() for i in range(10)]

/conda/envs/bsql/lib/python3.7/site-packages/pyblazing/apiv2/context.py in sql()
   1499 
   1500         if (algebra is None):
-> 1501             algebra = self.explain(sql)
   1502 
   1503         if self.dask_client is None:

/conda/envs/bsql/lib/python3.7/site-packages/pyblazing/apiv2/context.py in explain()
    954         Docs: https://docs.blazingdb.com/docs/explain
    955         """
--> 956         return str(self.generator.getRelationalAlgebraString(sql))
    957 
    958     def add_remove_table(self, tableName, addTable, table=None):

Exception: java.lang.IllegalStateException: Unable to instantiate java compiler
```

### Case 2
Code:
```
import cudf, glob
from blazingsql import BlazingContext
from dask.distributed import Client
from dask_cuda import LocalCUDACluster
from dask.distributed import Client
from dask.delayed import delayed
import dask

cluster = LocalCUDACluster()
client = Client(cluster)

bc = BlazingContext()
def ct(i):
    bc.create_table(f'tab{i}', cudf.DataFrame({'a':range(10000)}))
    
def fun(i):
    return bc.sql(f'select * from tab{i}')

[ct(i) for i in range(10)]
x = [delayed(fun)(i) for i in range(10)]
dask.compute(x)

```
Exception:
```
/conda/envs/bsql/lib/python3.7/pickle.py in save_dict(self, obj)
    857 
    858         self.memoize(obj)
[...]

PR #580, opened 2020-04-24
[Testing] added changes to be able to run a query on a single gpu via dask and …

felipeblazing:
…also to return a future instead of the materialized result set

resolves https://github.com/BlazingDB/blazingsql/issues/566

Basically what this pr does is add two parameters to .sql()

single_gpu = True|False (default False)
This allows the user to specify that the query in question will be run on a single gpu. Can be useful if a user has prepartitioned tables and wants to line them up to avoid shuffles during join.

return_futures = True|False (default False)
This allows the user to specify that instead of a dask dataframe they want to get a future to be able to get that dataframe later on. Makes .sql return almost instantly since now the query is happening asynchronously


williamBlazing:
This will be implemented in #610

PR #665, opened 2020-05-19
[TESTING] Detect worker crashed while running query

### LINKED ISSUE #629: Make it so that dask can handle failed futures and not crash the kernel
When a crash occurs how are we supposed to handle it? How do we cancle the rest of the query so that we don't hang forever?

### LINKED ISSUE #643: Fix issue hanging when worker crashes
I believe that most of this hanging occurs in ExternalBatchColumnDataSequence  in BatchProcessing.h . This issue deals with hanging only and is not yet resolving issues that arise from the fact that the new dask workers that have been restarted have not had their blazing context initialized yet. 

Basically what happens is as follows:

This is the loop that I think is the problem

```
BlazingMutableThread t([this, comms_message_token, context_token, message_id](){
			while(true){
					auto message = Server::getInstance().getHostMessage(context_token, comms_message_token);
					if(!message) {
						--last_message_counter;
						if (last_message_counter == 0 ){
							this->host_cache->finish();
							break;
						}
					}	else{
						auto concreteMessage = std::static_pointer_cast<ReceivedHostMessage>(message);
						assert(concreteMessage != nullptr);
						auto host_table = concreteMessage->releaseBlazingHostTable();
						host_table->setPartitionId(concreteMessage->getPartitionId());
						this->host_cache->addToCache(std::move(host_table), message_id, this->context.get());			
					}
			}
		});
```

1. while(true) is a NEVER EVER DO THIS 
instead auto message = Server::getInstance() can be block and have some condition variable inside it waiting for things to be added to the queue
2. we have no way of knowing if this message every will return. on solution is
Send a message from clieent to server with a random number
server stores number in a list
client heartbeats with that number to server
server checks if it has that number in its list, if it does returns success else returns fail
if client gets fail or gets no response it knows that the worker restarted or died
in this case it will never get a response so we should throw an exception

### LINKED ISSUE #644: Be able to recover cluster from a crash
Right now we hang when we crash. That is being dealt with in https://github.com/BlazingDB/blazingsql/issues/643 but in addition to this we want to make sure that the context that comes back after a crash is in a state that is predictable. To accomplish this there are a few things that we need to be able to do

1. stop storing any state about the Nodes in context. The information should be queried from the scheduler each time. If we find that the latency this add is unacceptable we can revisit but its much safer to ask the schdeduler what workers it has available

2. When crashes do occur the dask_cudf tables that have been created are not longer valid and need to be removed so they cant erroneously be used

3. Every time we run a query we should send some kindf of initalize_if_not_yet_initalized command to each dask-worker to make sure we are in a good state when we start running the query.

jeanp413:
Fixes #629 
Fixes #644
Fixes #643

After a crash use it like this to reset the cluster state
```
    try:
        gdf =  bc.sql(sql)
    except:
        bc.reset()
```


pyblazing/pyblazing/apiv2/context.py
```diff
@@ -1709,10 +1718,23 @@ def sql(self, query, table_list=[], algebra=None, return_futures=False, single_g
...
+                    # Check for crashed workers
+                    while any([not f.done() for f in dask_futures]):
+
+                        workers_info = self.dask_client.scheduler_info()["workers"]
+                        last_seen_times = [workers_info[worker]['last_seen'] for worker in workers_info]
+                        if any(time > start_query_time for time in last_seen_times):
+                            raise RuntimeError("A worker crashed during query execution")
+                            break
+
+                        time.sleep(0.5)
```
williamBlazing, line 1734: This sleep here will make it so that the fastest a query will return is 0.5 sec, followed by 1 sec, etc.... We should at the very least make this sleep much smaller like 0.1 or 0.2 sec, or figure out something other than a spin lock here

williamBlazing:
No longer going to be an issue with ucx

PR #751, opened 2020-06-05
[Testing] Fix/add error handling messaging tcp

felipeblazing:
This PR fixes a few things.
1. 
PinnedBufferProvider
 Changes cudaMallocManaged to cudaMallocHost in Pinned Buffers which was supposed to be a host allocated pinned memory buffer not uvm. Removed a bunch of code that existed due to this fact that basically deleted and recreated the entire BufferProvider which was both incredibly ineficcient and contrary to the idea of using a pool. 
Another problem with the way that it was working is that the provider was basically single thread access only safe. If two threads started using it at the same time they could invalidate all of their buffers since Free would have been called on all of them. Probably need to investigate more fully to make sure that the pinnedbuffer provider is shared. 

2. ZMQ exceptions where being caught with nothing in the catch scope. e.g. catch( ... ){ }
Made it so these errors propagate up properly. In order to do this safely without having to insert intermediate try catches and fianlly clauses that check to see which resources were freed up and which weren't I changed it so that the PinnedBuffer provider now gives us an object that lets us get at the data and size of the buffer like before but when it goes out of scope it makes the buffer available again to the pinnedBuffer provider instead of us having to manually call pinnedBufferProvider.freeBuffer() all over the place which was error prone.


felipeblazing:
Before we merge this in I am going to be running some performance tests to make sure we see no degradation

williamBlazing:
We wont need this since we are going to ucx

PR #753, opened 2020-06-05
[REVIEW] Setting host memory threshold parameter

### LINKED ISSUE #736: [BUG] Dask shuts down workers when host RAM is about to run out
**Describe the bug**
The engine is able to calculate when it is going to start moving frames from GPU memory to Host memory and then to disk. However, with very large data queries it is not working well since the workers are being shut down with the message below:

distributed.nanny - WARNING - Worker exceeded 95% memory budget. Restarting

Run a query with large data enough to use the disk after run out the host memory.

**Expected behavior**
It should work smoothly with the bigger than GPU architecture.

**Environment overview**
 - Method of cuDF install: conda
 - Method of blaazingsql install: from source, branch-0.14

**Additional context**
Apparently the engine should be considering this Dask feature to shutdown workers automatically (by default the threshold is set to 0.95).

https://github.com/dask/distributed/blob/d8f46d502a561bd86cc5c08eec0fd906a4ba700d/distributed/distributed.yaml#L80-L86

rommelDB:
Closes #736


engine/bsql_engine/io/cio.pxd
```diff
@@ -163,7 +163,7 @@ cdef extern from "../include/engine/engine.h":
         TableScanInfo getTableScanInfo(string logicalPlan)
 
 cdef extern from "../include/engine/initialize.h":
-    cdef void initialize(int ralId, int gpuId, string network_iface_name, string ralHost, int ralCommunicationPort, bool singleNode, map[string,string] config_options) except +raiseInitializeError
+    cdef void initialize(int ralId, int gpuId, string network_iface_name, string ralHost, int ralCommunicationPort, bool singleNode, map[string,string] config_options, float host_memory_quota) except +raiseInitializeError
```
williamBlazing, line 166: why are you sending `host_memory_quota` as a separate parameter when you could pass it in with all the other config_options?

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -830,6 +834,10 @@ def __init__(self, dask_client=None, network_interface=None, allocator="managed"
         for option in config_options:
             self.config_options[option.encode()] = str(config_options[option]).encode() # make sure all options are encoded strings
 
+        host_memory_quota = 0.95
```
williamBlazing, line 837: I would set the default to be lower than 0.95. I think 0.75 would be better. 
Also, I think we should assume that if a user is manually setting that, that they know that each worker will be using effectively the same pool of memory since its the same resource. We can help with that by adding a little more explanation in the comments about BLAZING_HOST_MEM_RESOURCE_CONSUMPTION_THRESHOLD.
Also, i think that the default should always be adjusted based on the number of GPUs per server. You have logic that already tries to do that, but I think it should be easy to extrapolate to even if its not just one machine. Assuming a default of 0.75, you can basically do:
0.75 * len(set([value['host'] for key,value in self.dask_client.scheduler_info()["workers"].items()])) /  len(self.dask_client.scheduler_info()["workers"])
Its not perfect, but in most situations you will have the same number of GPUs per server.

So in summary, lets make it so that if the user sets something, you leave it as it is. If the user does not provide a value, then set a value depending on the number of gpus per server

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -793,6 +794,9 @@ def __init__(self, dask_client=None, network_interface=None, allocator="managed"
                                     BLAZING_DEVICE_MEM_RESOURCE_CONSUMPTION_THRESHOLD : The percent (as a decimal) of total GPU memory that the memory resource
                                             will consider to be full
                                             default: 0.95
+                                    BLAZING_HOST_MEM_RESOURCE_CONSUMPTION_THRESHOLD : The percent (as a decimal) of total host memory that the memory resource
```
williamBlazing, line 797: We have the ability to set config_options during BlazingContext creation and also when running .sql(). We should mention in these comments that BLAZING_HOST_MEM_RESOURCE_CONSUMPTION_THRESHOLD and BLAZING_DEVICE_MEM_RESOURCE_CONSUMPTION_THRESHOLD can only be set in BlazingContext creation

rommelDB: Added endnote.

engine/bsql_engine/io/cio.pxd
rommelDB: I put this before considering the host threshold to be a config parameter, removed.

pyblazing/pyblazing/apiv2/context.py
rommelDB: Fixed. Please, check again.

engine/src/bmr/BlazingMemoryResource.h
```diff
@@ -276,15 +273,10 @@ class blazing_device_memory_resource : public BlazingMemoryResource {
 /**
 	@brief This class represents a custom host memory resource used in the cache system.
 */
-class blazing_host_memory_mesource : public BlazingMemoryResource{
+class internal_blazing_host_memory_resource{
```
felipeblazing, line 276: Why is there this seperate internal class?

PR #763, opened 2020-06-09
[REVIEW] implemented ability to set the folder for all log files

### LINKED ISSUE #710: Move all logs to a .blazing_log folder
It is incredibly cluttered when looking at the directory from which you are running blazing. Those logs should be put somewhere less obtrusive.

williamBlazing:
Resolves #710 

Tested with relative and absolute path
Tested setting the path in config_options and leaving it default
Tested in single node, and distributed, both with LocalCudaCluster and manually launching dask workers.

Have not tested with multiple servers.


williamBlazing:
rerun-tests

Issue #749, opened 2020-06-05
BlazingContext crash when we don't use any bc method


aucahuasi:
Run this single script 
```python
import blazingsql
bc = blazingsql.BlazingContext()
```
and there is a crash
```shell
listening: tcp://*:27247
BlazingContext ready
FATAL: exception not rethrown
Aborted
```

williamBlazing:
We should look into this AFTER UCXpy has been implemented

Christian8491:
Currently this behavior has disappeared. Just for let you know.

PR #775, opened 2020-06-09
[PAUSED] Call finalizeCaller in the destructor

Christian8491:
Closes partially  #749 

CASE 1. This works
```

def main():
	bc = BlazingContext()

if __name__ == '__main__':
	main()
```

CASE 2. This does not work
```

bc = BlazingContext()
```
The Python garbage collector is not working well in case 2. 
@aucahuasi  @gcca  do you  have thoughts for these cases?  If someone of you consider a better option, please let me know. cc @williamBlazing


williamBlazing:
implementation did not quite work. Ucx may fix underlying issues

PR #777, opened 2020-06-10
[REVIEW] Updating Calcite

### LINKED ISSUE #766: Update Calcite
A user is reporting an error coming from Calcite. Lets update Calcite to see if the error goes away

rommelDB:
Fixes #766


algebra/blazingdb-calcite-application/pom.xml
```diff
@@ -71,6 +71,13 @@
 			<artifactId>commons-cli</artifactId>
 			<version>1.4</version>
 		</dependency>
+
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-dbcp2</artifactId>
```
aucahuasi, line 77: Why we need this new dependency  @rommelDB ?

rommelDB: This dependency isn't new. We used to use it on https://github.com/BlazingDB/blazingsql/blob/branch-0.15/algebra/blazingdb-calcite-application/src/test/java/com/blazingdb/calcite/catalog/BlazingCatalogTest.java#L18 ..  difference came from Calcite 1.21, there the commons-dbcp2 package is a compile-time dependency, meanwhile, on Calcite 1.23 commons-dbcp2 is a runtime dependency and this caused a not found error.

PR #784, opened 2020-06-11
[REVIEW] Full TPCH coverage

### LINKED ISSUE #703: Evaluate missing features for full TPC-H coverage
In general the capabilities of our engine have gotten much better. We need see how many of the TPC-H queries we dont actually support and why. So far, it seems like there are only two features missing:
Inequality joins
Cartesian joins

To do a thorough evaluation, we need to revisit the `tpchQueriesTest.py` from the testing framework. We want to look at each query and make sure that its running the same way as in the official TPC-H query. If not, any necessary modifications need to be explicitly mentioned. If any modifications are necessary, the script should contain the original and the modified and why it was modified.

Once we have a new official set of queries for us to run, we need to see which ones pass and which ones dont. The ones that dont, we want to identify if its because of any of the two missing features listed above or if its something else.

kharoc:
Closes #703


GPUtester:

rommelDB:
add to whitelist

williamBlazing:
rerun tests

rommelDB:
rerun tests

williamBlazing:
rerun tests

williamBlazing:
add to whitelist

PR #786, opened 2020-06-12
[REVIEW] Detect string overflow

### LINKED ISSUE #761: Tpcx-bb Q19 issue
Q19: Dask-cudf part. RuntimeError: cuDF failure at: /conda/conda-bld/libcudf_1591170266934/work/cpp/src/strings/copying/concatenate.cu:232: total size of strings is too large for cudf column

rommelDB:
Fixes #761


engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
```diff
@@ -104,6 +104,14 @@ class PartwiseJoin : public kernel {
 			for (std::size_t i = 0; i < tables_loaded.size(); i++){
 				tables_to_concat[i] = tables_loaded[i]->toBlazingTableView();
 			}
+
+			if( ral::utilities::checkIfConcatenatingStringsWillOverflow(tables_to_concat)) {
```
williamBlazing, line 108: Aqui podemos controlar que tanto queremos concatenar, entonces no queremos lanzar un warning, sino que queremos poner `checkIfConcatenatingStringsWillOverflow` en la condicion del while loop que crea `tables_loaded`

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
```diff
@@ -474,6 +474,16 @@ std::unique_ptr<ral::frame::BlazingTable> ConcatenatingCacheMachine::pullFromCac
 			tables_holder[i] = std::move(data->decache());
 			table_views[i] = tables_holder[i]->toBlazingTableView();
 		}
+		if( ral::utilities::checkIfConcatenatingStringsWillOverflow(table_views)) {
```
williamBlazing, line 477: The concatenating cache can usually control how much to concatenate, so we want to add this check in the while loop that collects the messages to concatenate. 
For the case of single node processing, we always want to concatenate all. This is set in the creation of the concatenating cache for the final cache:
https://github.com/BlazingDB/blazingsql/blob/branch-0.15/engine/src/CalciteInterpreter.cpp#L98

Usually we would do this by leaving the `cache_settings` in its default setting, where it just reads everything.
https://github.com/BlazingDB/blazingsql/blob/branch-0.15/engine/src/execution_graph/logic_controllers/taskflow/kpair.h#L18

We add a new `cache_setting`: `bool concat_all` that is only for the Concatenating caches, that will force it to read all, regardless of the thresholds. And use that for the creation of the last cache.
https://github.com/BlazingDB/blazingsql/blob/branch-0.15/engine/src/CalciteInterpreter.cpp#L98

Use that new `concat_all` setting to be able to override `checkIfConcatenatingStringsWillOverflow` or the thresholds when collecting all the messages. And if it is going to concat_all and strings will overflow, then it should provide a warning

engine/src/execution_graph/logic_controllers/LogicalFilter.cpp
```diff
@@ -125,6 +125,11 @@ std::unique_ptr<ral::frame::BlazingTable> process_filter(
       }
     }
     assert(found_self_partition);
+
+		if( ral::utilities::checkIfConcatenatingStringsWillOverflow(partitions_to_concat)) {
+			std::cout<<"WARNING: In process_distribution_table Concatenating Strings will overflow strings length"<<std::endl;
```
williamBlazing, line 130: Have this std::cout be output to the logs. 
This function is called by bc.partition, which does not have a query id or anything like that, so you can leave all the logging fields blank, encept for the `info` where you provide this warning.
Also, since you are here, can you add a comment at the declaration of this function that this is only used by `bc.partition`

engine/src/io/DataLoader.cpp
```diff
@@ -150,6 +150,10 @@ std::unique_ptr<ral::frame::BlazingTable> data_loader::get_metadata(int offset)
 	if (metadata_batches.size() == 1){
 		return std::move(metadata_batches[0]);
 	} else {
+		if( ral::utilities::checkIfConcatenatingStringsWillOverflow(metadata_batche_views)) {
+			std::cout<<"WARNING: In data_loader::get_metadata Concatenating Strings will overflow strings length"<<std::endl;
```
williamBlazing, line 154: Have this std::cout be output to the logs.
You wont have a query id or anything like that, so you can leave all the logging fields blank, except for the info where you provide this warning.

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
```diff
@@ -474,6 +474,17 @@ std::unique_ptr<ral::frame::BlazingTable> ConcatenatingCacheMachine::pullFromCac
 			tables_holder[i] = std::move(data->decache());
 			table_views[i] = tables_holder[i]->toBlazingTableView();
 		}
+
+		if( concat_all && ral::utilities::checkIfConcatenatingStringsWillOverflow(table_views) ) {
```
williamBlazing, line 478: Here we are throwing a warning if concat_all is true and if concatenating strings will overflow. That is ok, to just throw a warning because we cant do anything about it. But if concat_all is false, and concatenating strings will overflow, then we do need to do something about it

williamBlazing:
rerun tests

williamBlazing:
rerun tests

Issue #758, opened 2020-06-08
Test what happens if you run a query that does literals only operations


williamBlazing:
Queries where one of the elements in the select does not actually use a data column.
I dont think this is currently supported, and should just return at least a good error message and not crash.

This task is to just test this scenario. If it does crash, then we should create a new issue

We want to test queries of this sort:

select 3 + 4, columnA from table where columnA = 2
select 3 + 4, 123 / 2 from table

williamBlazing:
Recent changes in the engine, in particular in Caclite have allowed us to now support operations on only literals.
We still want to create a suite of e2e tests that test that capability in different scenarios.
Lets test this with joins, groupbys, sorts and unions
Lets also test joins against literals:
inner join tableA.colA = 4
inner join tableA.colA = tableB.colA and tableA.colA = 5

Issue #451, opened 2020-03-09
[BUG] Inconsistent float expression evaluation in blazing


alexsisu:
**Describe the bug**
```
pdf = pd.DataFrame.from_dict({"a":[1.0,2.0]})
cdf = cudf.from_pandas(pdf)
bc.create_table("demo_table",cdf)
res = bc.sql("select a,(a-1.0)/(2.0-1.0) as x from demo_table")
print(res)
res = bc.sql("select a,a/(2.0-1.0) as y,(a-1.0)/(2.0-1.0) as x from demo_table")
print(res)
```

Output is:
```
     a             x
0  1.0  4.341048e-19
1  2.0  6.505213e-19
     a    y    x
0  1.0  1.0  0.0
1  2.0  2.0  0.0
```

Basically the result of `(a-1.0)/(2.0-1.0)` should be evaluated to 0 and 1. However, everything is evaluated to 0,0.

**Steps/Code to reproduce bug**
Run the above snippet

**Expected behavior**
Expected to be evaluated to [0,1]

**Environment details**
Printenv.sh output:
```
CONDA_SHLVL=2
LD_LIBRARY_PATH=/usr/local/cuda-10.0/lib64:
[...]

PR #802, opened 2020-06-17
[REVIEW] Feature/timestampadd and constant expressions evaluation by Calcite

### LINKED ISSUE #630: Implement TIMESTAMPDIFF and TIMESTAMPADD
This is the syntax expected by Calcite

{fn TIMESTAMPADD(timeUnit, count, datetime)} | Adds an interval of count timeUnits to a datetime

{fn TIMESTAMPDIFF(timeUnit, timestamp1, timestamp2)} | Subtracts timestamp1 from timestamp2 and returns the result in timeUnits

timeUnit is just a key set of possible keywords (SECOND, MINUTE, HOUR, DAY, MONTH, YEAR)

for TIMESTAMPADD:
- count should be an integer that is from a column or a literal
- datetime should be a timestamp type that is form a column or a literal

for TIMESTAMPDIFF:
- timestamp1 should be a timestamp type that is form a column or a literal
- timestamp2 should be a timestamp type that is form a column or a literal

NOTE: we may not be able to support operations on purely literals at this point

rommelDB:
Fixes #630 #758 #451


rommelDB:
rerun tests

rommelDB:
rerun tests

PR #821, opened 2020-06-22
[WIP] UCX PR

felipeblazing:
This PR will be the one that implements UCX communication in Blazing. So far here we have created Caches and bindings for them in Python that UCX can use for transmitting messages.


williamBlazing:
You should consider creating a set of Factories for building your ral::cache::MetadataDictionary metadata structures. Here are some of the benefits:
- A lot of the data that is sent uses very similar metadatas that are build almost the same way. If you have a factory for them you eliminate a lot of duplicated code. 
- Any new distributed algorithms can use the same factories and developers dont have to go figure out all the fields that need to be filed in and with what. 
-If we have a set of Factories, then we have all the types of messages that get sent in one place. Its easier to update how that code works if its more organized in one place. 
- If the metadata creation is just one line, its easier to read and understand what the code is doing with that metadata. Especially if the factory function name explains the message or data transmission that the metadata is for

felipeblazing:
That doesnt make it simpler in my mind. Basically instead of being clear that I am setting metadata I am calling some other function, have to go and look up the definition of that function to make sure I pass in the parameters in the right order. It really doesn't simplify anything it just makes the code shorter but at the cost of obfuscating what is actually happening. 

Instead of obfuscating whats happening I think we need to start using composition more. Have a DistributableKernel class that we can use to have methods that can help us generate this metadata and therefore we only have to set the values that are part of this kernel. In general we send three kinds of messages. 

1. Data alone
2. metadata alone where metadata can vary quite a bit in its contents
3. metadata and data where the metadata is standard

All of the context does not need to be passed to partition_table. It could be a method that exists in DistributableKernel and all the kernels that can be distributed implement it. Then the function would look something more like this.

```
    metadata.add_value(ral::cache::CACHE_ID_METADATA_LABEL, cache_id); //Because a kernel can have multiple
for (auto i = 0; i < partitions_to_send.size(); i++) {
    blazingdb::transport::Node dest_node;
    ral::frame::BlazingTableView table_view;
    std::tie(dest_node, table_view) = partitions_to_send[i];
    if(dest_node == self_node ) {
        continue;
    }

    metadata.add_value(ral::cache::WORKER_IDS_METADATA_LABEL, dest_node.id());
    this->send(table_view,metadata);

}

```

williamBlazing:
The work of this PR is continued here https://github.com/BlazingDB/blazingsql/pull/911

PR #854, opened 2020-07-06
[ENH] Added envvar to hints for CUDFTESTUTIL_LIBRARY for alternate direcotry of 'tests'

millerhooks:
Adjacent to #849, this makes sure the `cudftestutils` library is discoverable if the environment variable is already set.

### DIFF
--- a/conda-build-docker.sh
+++ b/conda-build-docker.sh
@@ -271,3 +271,4 @@ fi
 
 logger "You can run docker exec on the containers, if not needed then just kill them!"
 docker ps -a
+


millerhooks:
I thought this worked but it was some `ccache` from hard coding the path. Updating CMakeLists.txt with proper checks and variables and will push back shortly.

millerhooks:
Had some ccache values from hard coding the issue and this passed tests. Working out the updates to CMakeTests.txt to solve the problem, will submit update shortly.

millerhooks:
Looks like we drop it in `/tests` a lot. So I just added that to the hints. I don't know that it needs to be more robust than that.

millerhooks:
Sorry. Doing nearly a CI process for rebuild. Pushed a forced commit and wrecked it. I'll push a new PR when I've got it worked out.

PR #860, opened 2020-07-08
[ENH] Added path to CMakeList.txt cudftestutil to look in test for lib

millerhooks:
With #849 in some situations `cudftestutil` is built in `test` instead of `lib` adding that path to library `HINTS`.


comms/tests/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ function(configure_test TEST_NAME Tests_SRCS)
     set(CUDF_TEST_INCLUDE "$ENV{CUDF_HOME}/cpp")
 
     find_library(CUDFTESTUTIL_LIBRARY "libcudftestutil.a"
-                HINTS "$ENV{CUDF_ROOT}/lib")
+                HINTS "$ENV{CUDF_ROOT}/lib" "$ENV{CUDF_ROOT}/tests")
```
aucahuasi, line 15: What is the env var CUDF_ROOT and when is defined and by what process? ... is the same as our env var CUDF_HOME?

Issue #842, opened 2020-06-30
Removing global locks when calling pure C functions


williamBlazing:

felipeblazing:
The look at the code in https://github.com/BlazingDB/blazingsql/blob/branch-0.15/engine/bsql_engine/io/io.pyx and find all the functions that are cdef only. Those functions are calling only c apis. Any code that calls c apis should be inside of a 
with nogil: 

and nogil must be added to the c api that it is calling defined in 
https://github.com/BlazingDB/blazingsql/blob/branch-0.15/engine/bsql_engine/io/cio.pxd

None of our c apis modify python classes in a non thread safe manner so we can always release the gil.

PR #866, opened 2020-07-08
Added nogil statements for pure C functions in Cython

rlratzel:
Added nogil statements for pure C functions in Cython.

I think this is correct based on Cython compiling cleanly and the tests passing, but I never actually benchmarked a before-and-after run to confirm a speedup.

closes https://github.com/BlazingDB/blazingsql/issues/842


engine/bsql_engine/io/cio.pxd
```diff
@@ -144,29 +144,28 @@ cdef extern from * namespace "blazing":
         template <class T> inline typename std::remove_reference<T>::type&& blaz_move(T&& t) { return std::move(t); }
         }
         """
-        cdef T blaz_move[T](T)
+        cdef T blaz_move[T](T) nogil
```
felipeblazing, line 147: this is a pretty trivial function and the overhead of releasing and reacquiring the gil is probably more than just allowing it to use the gil while calling move

rlratzel: Yes I can remove that. FYI, a `nogil` function annotation doesn't actually release the GIL, it just doesn't attempt to acquire it: https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html?highlight=nogil#declaring-a-function-as-callable-without-the-gil
```
The nogil function annotation declares that it is safe to call the function without the GIL. It 
is perfectly allowed to execute it while holding the GIL. The function does not in itself release 
the GIL if it is held by the caller. 
```
Is that helpful or would you still prefer it be removed?

rlratzel:
> These are the functions that we want to call with nogil thats looking good. Can you look up their invocations which are in io.pyx and wrap them in
> 
> ```
> with nogil:
>     ...
> ```

Yes I misunderstood what the `nogil` function declaration would do. It won't actually release the GIL - which is what you want - instead it just won't attempt to acquire.  I'll fix that.

engine/bsql_engine/io/cio.pxd
rlratzel: Turns out `blaz_move()` has to be declared with `nogil` if it's to be called from within a `with nogil ...` block.

felipeblazing: That's fine leave it with nogil

rlratzel:
I just made a commit to use a pattern recommended by @shwina which cudf also uses to mark functions included from a header file as `nogil` safe, instead of declaring every individual function.

rlratzel:
rerun tests

rlratzel:
rerun tests

rlratzel:
rerun tests

PR #879, opened 2020-07-14
[REVIEW] Fix log directory creation

### LINKED ISSUE #856: [BUG] BlazingContext with multi-node dask cluster requires ~/blazing_log to exist on all nodes
When trying to set up a BlazingContext with a multi-node Dask cluster, I get an error if `~/blazing_log/` does not exist on every node. It appears that it wants to write logs into that directory, but cannot. Perhaps this could be created automatically, maybe by utilizing the passed `client`?

```

bc = BlazingContext(
        dask_client=client, # multi-node cluster client
        pool=True,
        network_interface=os.environ.get("INTERFACE", "eth0"), # INTERFACE=ib0
    )

Traceback (most recent call last):
  File "tpcx_bb_query_07_sql.py", line 84, in <module>
    network_interface=os.environ.get("INTERFACE", "eth0"),
  File "/home/nicholasb/conda/envs/rapids-tpcxbb-20200707-bsql/lib/python3.7/site-packages/pyblazing/apiv2/context.py", line 952, in __init__
    ralPort, ralIp, log_path = connection.result()
  File "/home/nicholasb/conda/envs/rapids-tpcxbb-20200707-bsql/lib/python3.7/site-packages/distributed/client.py", line 223, in result
    raise exc.with_traceback(tb)
  File "/home/nicholasb/conda/envs/rapids-tpcxbb-20200707-bsql/lib/python3.7/site-packages/pyblazing/apiv2/context.py", line 109, in initializeBlazing
    logging.basicConfig(filename=filename,format=FORMAT, level=logging.INFO)
  File "/home/nicholasb/conda/envs/rapids-tpcxbb-20200707-bsql/lib/python3.7/logging/__init__.py", line 1895, in basicConfig
    h = FileHandler(filename, mode)
  File "/home/nicholasb/conda/envs/rapids-tpcxbb-20200707-bsql/lib/python3.7/logging/__init__.py", line 1087, in __init__
    StreamHandler.__init__(self, self._open())
  File "/home/nicholasb/conda/envs/rapids-tpcxbb-20200707-bsql/lib/python3.7/logging/__init__.py", line 1116, in _open
    return open(self.baseFilename, self.mode, encoding=self.encoding)
FileNotFoundError: [Errno 2] No such file or directory: '/home/nicholasb/blazing_log/pyblazing.8.log'
```

From the 2020-07-07 nightly around 11 AM EDT.
```
blazingsql                0.15.0a         cuda10.2_py37_262    blazingsql-nightly/label/cuda10.2
```

rommelDB:
Fixes #856


pyblazing/pyblazing/apiv2/context.py
```diff
@@ -681,36 +681,57 @@ def adjust_due_missing_rowgroups(metadata, files):
...
-        initialized[host] = False
 
     dask_futures = []
-    for worker, worker_info in client.scheduler_info()["workers"].items():
-        if not initialized[worker_info["host"]]:
-            dask_futures.append(
-                client.submit(initialize_server_directory, dir_path, workers=[worker])
-            )
-            initialized[worker_info["host"]] = True
+    i = 0  # this extra index is to avoid scheduler optimization
```
williamBlazing, line 689: A better way of doing this (that we learned recently) is to do like this:
```
for worker, worker_info in all_items:
            dask_futures.append(	        
                client.submit(initialize_server_directory, dir_path, workers=[worker], pure=False)	            
            )	        
```
Using the parameter `pure=False` makes it so that you dont have to pass it that `i`

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -681,36 +681,57 @@ def adjust_due_missing_rowgroups(metadata, files):
...
+
+    current_dir_hosts = client.gather(dask_futures)
+
+    # lets map the working directories shared by several workers,
+    # in such a way that only one worker will create the logging folder
+    map_host_path = {}
+    for path, worker in zip(
+        current_dir_hosts, [worker for worker, worker_info in all_items]
+    ):
+        map_host_path[path] = worker
```
williamBlazing, line 704: I dont think this will work. If all workers have the same current working directory, then your `map_host_path` will only have one entry, which means only one of the workers will get initialized

williamBlazing: Additionally, this is only necessary if clients have different CWD AND if the `dir_path` is a relative path. If its an absolute path, it does not matter what the CWD is. Which means what you need to get is a set of  unique worker IP and actual resolved `dir_path`. By resolved, i mean if its a relative path, then "resolve" it against the CWD, if its absolute, then its already resolved.

pyblazing/pyblazing/apiv2/context.py
rommelDB: Fixed, works perfectly, thanks!

pyblazing/pyblazing/apiv2/context.py
rommelDB: You're right @williamBlazing, I didn't consider these scenarios. Thanks, I just improved the solution.

Issue #869, opened 2020-07-10
Separate WaitingQueue and write unit tests for it. 


felipeblazing:
Seperate waiting queue into its own header file and write unit tests for the methods in it. Make the message_ptr type that it contains be a templated type so that we can be able to change it for a mock type if we want to in the future.  After that we want to write unit tests for each of the functions found inside waiting queue. Make sure to include it in CacheMachine.h after and add teh default templated type for message ptr in CacheMachine.h

The class can be found https://github.com/BlazingDB/blazingsql/blob/branch-0.15/engine/src/execution_graph/logic_controllers/CacheMachine.h#L167

In general the purpose of this class is to add some thread safety to a Dequeue which manages CacheData objects, the objects referred to by message_ptr in the WaitingQueue class. We want to write some basic tests that just push and pull into them and make sure that the information is not being lost.

We should write unit tests for the functions:

Note*** If a function has the word wait in it these all have timeouts that basically log a warning that its taking a long time. There is also other logic here so we want to make sure that all of the tests that have a wait have a version that tests with both a short and long delay. 

put
- run this function from multiple threads using the same object and ensure that at then end the queue contains everything that was expected

pop_or_wait
- run this from multiple threads on a queue that is already filled up ensuring that you retrieved everything that started out in the queue
- start out with a  queue with some message_ptr already in it, run pop_or wait on multiple threads and have another thread which is adding more data to the queue while the other threads are pulling. You can add the messages by either using the put or somehow manually changing the contents of the dequeue in a lock safe way (using put is easier :) )
  - you should make this with both a short and long delay when it comes to adding messages to the queue so that at least once it has to wait for longer than the timeout of 60s

wait_for_next
- start out with an empty dequeue. call wait_for_next on n threads. Add n items to the dequeue (probably using put) and verify that all of the wait_for_next threads complete and have one of the n messages you inserted
  - make a version of this where the put starts after 60s to make sure it covers the timeout code section

wait_until_finished
- start a few threads which call wait until finished. On a second thread block execution for a few seconds then call the finish() function on WaitingQueue.
  - make a version where the second thread waits at least 60s before calling finish

pop_unsafe
 - start out with n messages in the dequeue and call pop_unsafe n times and make sure that you have the same messages back in a fifo order

get_all_or_wait
- start out with a queue that is empty. on one thread add n messages to it then call on the WaitingQueue finish(). On another thread call get_all_or_wait()
  - make another version of this where you wait for at least 60s before you call finish()
[...]

Issue #884, opened 2020-07-17
[BUG] WaitingQueue::get_or_wait() infinite loop if message_id never found


rlratzel:
While writing `WaitingQueue` unit tests, a condition was discovered that could be a potential infinite loop.
```
        message_ptr get_or_wait(std::string message_id) {
                CodeTimer blazing_timer;
                std::unique_lock<std::mutex> lock(mutex_);
                while(!condition_variable_.wait_for(lock, 60000ms, [message_id, &blazing_timer, this] {
                                auto result = std::any_of(this->message_queue_.cbegin(),
                                                        this->message_queue_.cend(), [&](auto &e) {
                                                                return e->get_message_id() == message_id;
                                                        });
                                bool done_waiting = this->finished.load(std::memory_order_seq_cst) or result;
                                if (!done_waiting && blazing_timer.elapsed_time() > 59000){
                                        // log stuff
                                }
                                return done_waiting;
                        })){}
                if(this->message_queue_.size() == 0) {
                        return nullptr;
                }
                while (true){
                        auto data = this->pop_unsafe();
                        if (data->get_message_id() == message_id){
                                return std::move(data);
                        } else {
                                putWaitingQueue(std::move(data));
                        }
                }
        }
```
The first `while` loop can complete if `WaitingQueue::finish()` is called, at which point the `while (true)` loop can start if `message_queue` is not empty.  If `message_id` is never found, the method will never return and cannot be interrupted externally.

One proposal is to add a mechanism to allow clients to stop all `get_or_wait()` calls (similar to calling `finish()`) during a shutdown.

**Steps/Code to reproduce bug**
* Add a message to the WaitingQueue using `put()`
* Call `get_or_wait()` with a msg ID that is not present in the queue
* In a separate thread, call `finish()`
* Observe that the call to `get_or_wait()` will never return.

**Expected behavior**
* While `finish()` may not be the appropriate API to allow `get_or_wait()` to return, some other API is needed to unblock clients waiting on messages that will never appear.

williamBlazing:
Another solution is to reacquire the lock right before calling:
```
if(this->message_queue_.size() == 0) {
                        return nullptr;
                }
                while (true){
                        auto data = this->pop_unsafe();
                        if (data->get_message_id() == message_id){
                                return std::move(data);
                        } else {
                                putWaitingQueue(std::move(data));
                        }
                }```

PR #885, opened 2020-07-17
[ENH] Added initial set of unit tests for WaitingQueue and nullptr checks around spdlog calls

rlratzel:
Added initial set of unit tests for `WaitingQueue` and nullptr checks around spdlog calls.

Also added a `nullptr` check to various calls to `spdlog` objects since the tests do not set up `spdlog`.  We may want to replace `spdlog` calls with calls to an abstract logger that can be injected into the `WaitingQueue` ctor. This will allow unit tests to pass in a logging mock that can be checked for correct content as part of a test.

Notes:
* Sorry for the noisy diff! My editor removed trailing whitespace.
* <s>There is not full coverage of the public API for this class yet, this is still a WIP. Descriptions of the tests that need to be added are [here](https://github.com/BlazingDB/blazingsql/issues/869)</s>
* These tests are not automatically callled, like others are, when `build.sh` is invoked. The gtest binary has to be run manually to run these tests.
* There's a helper function to create message objs to add to the queue as part of the tests. This helper could be updated to use mock objects if we have the need to do a deeper inspection on how `WaitingQueue` interacts with the objects, or if we need to completely decouple these tests from the other objects.
* I'm assuming the `nullptr` checks around the `spdlog` calls are a good safety measure, but maybe there's a nicer way to make the logging calls (eg. a macro) that's less verbose?  This could also be addressed by the logging abstraction proposal above which could decouple `WaitingQueue` from `spdlog` completely, and encapsulate checks like that in one place.
* Filed this issue: https://github.com/BlazingDB/blazingsql/issues/884

closes https://github.com/BlazingDB/blazingsql/issues/869


felipeblazing:
The macro is a good idea I filed an issue for this here https://github.com/BlazingDB/blazingsql/issues/893

PR #887, opened 2020-07-20
[REVIEW] concatenating cache improvement

williamBlazing:
This PR brings over the CacheMachine code from the UCX branch.
It also improves ConcatenatingCache::pullFromCache to not decache until necessary
It also replaces PartwiseJoin::load_set with a concatenating cache on each input of PartwiseJoin
It also gets rid of flow control batch count, so that flow control feature only now works using bytes count. (with default parameters that was not being used anyways). This was removed to reduce code complexity.

### DIFF
--- a/engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
+++ b/engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
@@ -76,9 +76,9 @@ public:
 		this->condition = get_named_expression(new_join_statement, "condition");
 		this->join_type = get_named_expression(new_join_statement, "joinType");
 
-		if (condition == "true") {
+		if (this->condition == "true") {
 			this->join_type = CROSS_JOIN;
-		}
+		}		
 	}
 
 	bool can_you_throttle_my_input() {
@@ -468,12 +468,12 @@ public:
 		split_inequality_join_into_join_and_filter(this->expression, new_join_statement, filter_statement);
 
 		// Getting the condition and type of join
-		std::string condition = get_named_expression(new_join_statement, "condition");
+		this->condition = get_named_expression(new_join_statement, "condition");
 		this->join_type = get_named_expression(new_join_statement, "joinType");
 
-		if (condition == "true") {
+		if (this->condition == "true") {
 			this->join_type = CROSS_JOIN;
-		}
+		}		
 	}
 
 	bool can_you_throttle_my_input() {
--- a/engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
+++ b/engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
@@ -66,6 +66,19 @@ public:
 
 		this->leftArrayCache = 	ral::cache::create_cache_machine(cache_machine_config);
 		this->rightArrayCache = ral::cache::create_cache_machine(cache_machine_config);
+
+		// lets parse part of the expression here, because we need the joinType before we load
+		std::string new_join_statement;
+		StringUtil::findAndReplaceAll(this->expression, "IS NOT DISTINCT FROM", "=");
+		split_inequality_join_into_join_and_filter(this->expression, new_join_statement, filter_statement);
+
+		// Getting the condition and type of join
+		this->condition = get_named_expression(new_join_statement, "condition");
+		this->join_type = get_named_expression(new_join_statement, "joinType");
+
+		if (condition == "true") {
+			this->join_type = CROSS_JOIN;
+		}
 	}
 
 	bool can_you_throttle_my_input() {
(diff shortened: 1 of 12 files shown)


williamBlazing:
rerun tests

williamBlazing:
rerun tests

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
```diff
@@ -75,66 +88,13 @@ class PartwiseJoin : public kernel {
...
-
-
     std::unique_ptr<ral::frame::BlazingTable> load_left_set(){
 
 		this->max_left_ind++;
-		// need to load all the left side, only if doing a FULL OUTER JOIN
-		bool load_all = this->join_type == OUTER_JOIN ?  true : false;
-		auto table = load_set(this->left_sequence, this->input_.get_cache("input_a"), load_all);
+		std::unique_ptr<ral::frame::BlazingTable> table;
+		if (this->left_sequence.wait_for_next()){
```
felipeblazing, line 95: why isnt this one operation like get_or_wait?

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
```diff
@@ -146,11 +106,12 @@ class PartwiseJoin : public kernel {
...
-		// need to load all the right side, if doing any type of outer join
-		bool load_all = this->join_type != INNER_JOIN ?  true : false;
-		auto table = load_set(this->right_sequence, this->input_.get_cache("input_b"), load_all);
+		std::unique_ptr<ral::frame::BlazingTable> table;
+		if (this->right_sequence.wait_for_next()){
+			table = this->right_sequence.next();
+		}
 		if (not right_schema && table != nullptr) {
-			right_schema = std::make_unique<TableSchema>(table->get_schema(),  table->names());
+			left_schema = std::make_unique<TableSchema>(table->get_schema(),  table->names());
```
felipeblazing, line 114: this should be right_schema

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
```diff
@@ -75,66 +88,13 @@ class PartwiseJoin : public kernel {
...
-		// This actually does not make it faster, because it makes it so that there are more chunks to do pairwise joins and therefore more join operations
-		// We may want to revisit or rethink this
-
-		while ((input.wait_for_next() && bytes_loaded < join_partition_size_theshold) ||
-					(load_all && input.wait_for_next())) {
-			tables_loaded.emplace_back(std::move(input.next()));
-			bytes_loaded += tables_loaded.back()->sizeInBytes();
-
-			//If the concatenation so far produces a string overflow, we put back the batch because in this case the order of the elements does not matter
-			if(!load_all && ral::utilities::checkIfConcatenatingStringsWillOverflow(tables_loaded)){
```
felipeblazing, line 100: where does this part happen now

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
```diff
@@ -502,6 +461,19 @@ class JoinPartitionKernel : public kernel {
 		this->query_graph = query_graph;
 		this->input_.add_port("input_a", "input_b");
 		this->output_.add_port("output_a", "output_b");
+
+		// lets parse part of the expression here, because we need the joinType before we load
+		std::string new_join_statement;
```
felipeblazing, line 466: you can consider making JoinRelatedKernels inherit from the same kernel interface to avoid having this initialization step in multiple places.

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
```diff
@@ -200,7 +207,7 @@ void CacheMachine::addCacheData(std::unique_ptr<ral::cache::CacheData> cache_dat
 		int cacheIndex = 0;
 		while(cacheIndex < this->memory_resources.size()) {
 			auto memory_to_use = (this->memory_resources[cacheIndex]->get_memory_used() + cache_data->sizeInBytes());
-			if( memory_to_use < this->memory_resources[cacheIndex]->get_memory_limit()) {
+			if( memory_to_use < 999999999999) {
```
felipeblazing, line 210: lets fix this little number which i was using to ensure i wasnt moving data when testing

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
```diff
@@ -251,11 +258,14 @@ void CacheMachine::addCacheData(std::unique_ptr<ral::cache::CacheData> cache_dat
...
 }
 
-void CacheMachine::addToCache(std::unique_ptr<ral::frame::BlazingTable> table, const std::string & message_id) {
-
+void CacheMachine::addToCache(std::unique_ptr<ral::frame::BlazingTable> table, const std::string & message_id, bool always_add) {
 	// we dont want to add empty tables to a cache, unless we have never added anything
-	if (!this->something_added || table->num_rows() > 0){
+	if (!this->something_added || table->num_rows() > 0|| always_add){
 		for (auto col_ind = 0; col_ind < table->num_columns(); col_ind++){
+			if (table->view().column(col_ind).offset() > 0){
```
felipeblazing, line 265: this was one of my changes do we really need it? I am not sure we do. its basically a no op if nothing needs to be done

engine/src/execution_graph/logic_controllers/CacheMachine.h
```diff
@@ -192,11 +285,26 @@ class WaitingQueue {
...
 	}
 
+
+
+	void wait_for_count(int count){
+
+		std::unique_lock<std::mutex> lock(mutex_);
+		condition_variable_.wait(lock, [&, this] () {
+			std::cout<<"message queue size is "<<this->processed<<std::endl;
+			return count <= this->processed;
```
felipeblazing, line 295: so this is an interesting isituation. It should really be == and if its ever > we should through an exception. not sure what you think

engine/src/execution_graph/logic_controllers/PhysicalPlanGenerator.h
```diff
@@ -314,8 +300,28 @@ struct tree_processor {
 			} else {
 				auto child_kernel_type = child->kernel_unit->get_type_id();
 				auto parent_kernel_type = parent->kernel_unit->get_type_id();
-				if ((child_kernel_type == kernel_type::JoinPartitionKernel && parent_kernel_type == kernel_type::PartwiseJoinKernel)
-					    || (child_kernel_type == kernel_type::SortAndSampleKernel &&	parent_kernel_type == kernel_type::PartitionKernel)
+				if (child_kernel_type == kernel_type::JoinPartitionKernel && parent_kernel_type == kernel_type::PartwiseJoinKernel) {
+
+					std::size_t max_data_load_concat_cache_bytes_size = 400000000; // 400 MB
```
felipeblazing, line 305: shouldnt this be configurable?

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
williamBlazing: This is now handled by the concatenating cache. This here was effectively doing the same as a concatenating cache. I made it so that now the inputs to a join are concatenating caches. This way, we have less redundant code and also a join can operate on a more traditional, simpler pullFromCache, instead of all this logic

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
williamBlazing: good point. Let fix that.

williamBlazing: well, btw get_or_wait is for array like access. I am just going to do `next()` since it will wait or return null if there are none

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
williamBlazing: OH SHIT good catch! I swear I double checked this. Triple checking is better

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
williamBlazing: I do agree its super redundant, but I will put it into a utility function. I dont like the idea of making an interface just for one utility function

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
williamBlazing: yeah, we are ensuring ownership lower down anyways

engine/src/execution_graph/logic_controllers/CacheMachine.h
williamBlazing: I agree on both counts, but this needs to be on our branch, because this function is not used here

engine/src/execution_graph/logic_controllers/PhysicalPlanGenerator.h
williamBlazing: It is, its JOIN_PARTITION_SIZE_THRESHOLD

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
felipeblazing: its to keep it RAII like the class that constructs should do the initializing of thigns as a general principle

williamBlazing: Yeah, and things are getting initialized in the constructor

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
williamBlazing: fixed

PR #891, opened 2020-07-24
[REVIEW] fixed issue caused by replacing join load_set with concatenating cache

williamBlazing:
fixed issue caused by replacing join load_set with concatenating cache and issue in distributing CROSS_JOIN

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -18,10 +18,11 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean thirdparty io comms libengine engine pyblazing algebra -t -v -g -n -h"
+VALIDARGS="clean update thirdparty io comms libengine engine pyblazing algebra -t -v -g -n -h"
 HELP="$0 [-v] [-g] [-n] [-h] [-t]
    clean        - remove all existing build artifacts and configuration (start
                   over) Use 'clean thirdparty' to delete thirdparty folder
+   update       - update cudf thirdparty code and update cudf conda packages
    thirdparty   - build the Thirdparty C++ code only
    io           - build the IO C++ code only
    comms        - build the communications C++ code only
@@ -86,6 +87,10 @@ if (( ${NUMARGS} != 0 )); then
     done
 fi
 
+# Get version number
+export GIT_DESCRIBE_TAG=`git describe --tags`
+export MINOR_VERSION=`echo $GIT_DESCRIBE_TAG | grep -o -E '([0-9]+\.[0-9]+)'`
+
 # Process flags
 if hasArg -v; then
     VERBOSE=1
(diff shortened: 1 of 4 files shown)


build.sh
```diff
@@ -137,7 +137,7 @@ if buildAll || hasArg io || hasArg libengine || hasArg thirdparty; then
             cmake -DCMAKE_CXX11_ABI=ON ..
         else
             cd ${REPODIR}/thirdparty/cudf
-            git pull
+            # git pull
```
felipeblazing, line 140: so theres two sides to this sword. if you just comment it out. we make it less likely to detect when these changes WILL break things. While this is still an issue becuase we are using unexposed classes form CUDF we should have some kind of flag that we can set that decides whether or not it will update that. If it will update the cudf from source then you need to update the cudf in your conda as well to match. If you don't you have mismatch errors which are VERY confusing for new developers to blazing and has caused people to lose hours trying to build.

engine/src/execution_graph/logic_controllers/PhysicalPlanGenerator.h
```diff
@@ -285,11 +285,32 @@ struct tree_processor {
...
+			auto parent_kernel_type = parent->kernel_unit->get_type_id();
 			if (children.size() > 1) {
 				char index_char = 'a' + index;
 				port_name = std::string("input_");
 				port_name.push_back(index_char);
-				if (parent->kernel_unit->can_you_throttle_my_input()){
+
+				if (parent_kernel_type == kernel_type::PartwiseJoinKernel) {
+
+					std::size_t join_partition_size_thresh = 400000000; // 400 MB
```
felipeblazing, line 297: Can you wrap this kind of funcionatily in a macro. This could very easily be a 1 line macro that is something like
BLAZ_GET_CONFIG( SOME_CONSTANT_NAME).

engine/src/execution_graph/logic_controllers/PhysicalPlanGenerator.h
```diff
@@ -298,8 +319,7 @@ struct tree_processor {
 					query_graph += link(*child->kernel_unit, (*parent->kernel_unit)[port_name], cache_machine_config);
 				}
 			} else {
-				auto child_kernel_type = child->kernel_unit->get_type_id();
-				auto parent_kernel_type = parent->kernel_unit->get_type_id();
+				
 				if (child_kernel_type == kernel_type::JoinPartitionKernel && parent_kernel_type == kernel_type::PartwiseJoinKernel) {
 
 					std::size_t join_partition_size_thresh = 400000000; // 400 MB
```
felipeblazing, line 325: could also benefit from the use of the macro i describe above

engine/src/execution_graph/logic_controllers/PhysicalPlanGenerator.h
williamBlazing: Lets improve how we manage the config parameters in a different PR

PR #892, opened 2020-07-24
`std::endl` to `'\n'`

servusDei2018:
Changed prominent occurrences of `std::endl` to `'\n'`for speed boost. `endl` includes unnecessary `std::flush`. Benchmarks show improved performance.

### DIFF
--- a/comms/src/blazingdb/transport/Client.cc
+++ b/comms/src/blazingdb/transport/Client.cc
@@ -61,7 +61,7 @@ public:
     zmq::message_t local_message;
     auto success = socket_ptr->recv(local_message);
     if (success.value() == false || local_message.size() == 0) {
-      std::cerr << "Client:   throw zmq::error_t()" << std::endl;
+      std::cerr << "Client:   throw zmq::error_t()" << '\n';
       throw zmq::error_t();
     }
 
@@ -113,7 +113,7 @@ public:
     zmq::message_t local_message;
     auto success = socket_ptr->recv(local_message);
     if (success.value() == false || local_message.size() == 0) {
-      std::cerr << "Client:   throw zmq::error_t()" << std::endl;
+      std::cerr << "Client:   throw zmq::error_t()" << '\n';
       throw zmq::error_t();
     }
 
--- a/comms/src/blazingdb/transport/Server.cc
+++ b/comms/src/blazingdb/transport/Server.cc
@@ -112,7 +112,7 @@ public:
 		try {
 			thread.join();
 		} catch(std::exception & e) {
-			std::cout << "ServerTCP::exception: " << e.what() << std::endl;
+			std::cout << "ServerTCP::exception: " << e.what() << '\n';
 		}
 	}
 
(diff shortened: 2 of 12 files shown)


GPUtester:

servusDei2018:
There are more occurrences but these are just the most prominent ones.

io/src/Library/Logging/CoutOutput.cpp
```diff
@@ -10,12 +10,12 @@ CoutOutput::~CoutOutput() {}
 
 void CoutOutput::flush(std::string && log) {
 	std::unique_lock<std::mutex> lock(mutex);
-	std::cout << log << std::endl;
+	std::cout << log << '\n';
```
felipeblazing, line 13: the point of this function is to flush. It absolutely wants to use std::endl

io/src/Library/Logging/FileOutput.cpp
```diff
@@ -29,19 +29,19 @@ FileOutput::~FileOutput() {}
 
 void FileOutput::flush(std::string && log) {
 	std::unique_lock<std::mutex> lock(mutex);
-	file << log << std::endl;
+	file << log << '\n';
```
felipeblazing, line 32: this function is meant to flush.

io/src/Library/Logging/FileOutput.cpp
```diff
@@ -29,19 +29,19 @@ FileOutput::~FileOutput() {}
...
 }
 
 void FileOutput::flush(const std::string & log) {
 	std::unique_lock<std::mutex> lock(mutex);
-	file << log << std::endl;
+	file << log << '\n';
```
felipeblazing, line 37: thiose one is meant to flush as well

io/src/Library/Logging/FileOutput.cpp
```diff
@@ -29,19 +29,19 @@ FileOutput::~FileOutput() {}
...
 }
 
 
 void FileOutput::flush(
 	const int nodeInd, const std::string & datetime, const std::string & level, const std::string & log) {
 	std::unique_lock<std::mutex> lock(mutex);
-	file << datetime << "|" << nodeInd << "|" << level << "|" << log << std::endl;
+	file << datetime << "|" << nodeInd << "|" << level << "|" << log << '\n';
```
felipeblazing, line 44: another flusher!

comms/src/blazingdb/transport/Client.cc
```diff
@@ -61,7 +61,7 @@ class ConcreteClientTCP : public ClientTCP {
     zmq::message_t local_message;
     auto success = socket_ptr->recv(local_message);
     if (success.value() == false || local_message.size() == 0) {
-      std::cerr << "Client:   throw zmq::error_t()" << std::endl;
+      std::cerr << "Client:   throw zmq::error_t()" << '\n';
```
felipeblazing, line 64: If the concern is the perf cost of the flush. we might as well make this 
```

std::cerr << "Client:   throw zmq::error_t()\n;
```

comms/src/blazingdb/transport/Server.cc
```diff
@@ -216,7 +216,7 @@ void ServerTCP::Run() {
 					this->putMessage(message->metadata().contextToken, message);
 				}
 			} catch(const std::runtime_error & exception) {
-				std::cerr << "[ERROR] " << exception.what() << std::endl;
+				std::cerr << "[ERROR] " << exception.what() << '\n';
```
felipeblazing, line 219: cerr doesnt get buffered so I am not sure removing the flush here actually helps.

felipeblazing:
Some of the comments were resolved but the issues don't seem to be addressed. The functions that are called flush() are meant to flush the log.

williamBlazing:
ok to test

PR #894, opened 2020-07-27
added using exhale to be able to generate docs,

felipeblazing:
requires that you install doxygen apt-get also have to install exhale along with sphinx using pip


felipeblazing:
rerun tests

felipeblazing:
rerun tests

PR #914, opened 2020-07-31
[WIP] Implements run_batch

williamBlazing:


engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
```diff
@@ -25,30 +25,58 @@ class ComputeAggregateKernel : public kernel {
...
         bool ordered = false; // If we start using sort based aggregations this may need to change
-        BatchSequence input(this->input_cache(), this, ordered);
+        BatchSequenceBypass input(this->input_cache(), this); // WSM TODO, need to make BatchSequenceBypass take in "ordered" parameter
         int batch_count = 0;
         while (input.wait_for_next()) {
 
-            this->output_cache()->wait_if_cache_is_saturated();
-            auto batch = input.next();
+            // this->output_cache()->wait_if_cache_is_saturated();
+            std::vector<std::unique_ptr<ral::cache::CacheData>> batches;
```
felipeblazing, line 50: If i wanted to send this batch to multiple destinations what changes would have to be made to this code? I think that it almost cant be a unique_ptr

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
```diff
@@ -25,30 +25,58 @@ class ComputeAggregateKernel : public kernel {
...
+                    "step"_a=context->getQueryStep(),
+                    "substep"_a=context->getQuerySubstep(),
+                    "info"_a="ComputeAggregate Kernel Completed",
+                    "duration"_a=timer.elapsed_time(),
+                    "kernel_id"_a=this->get_id());
+
+        return kstatus::proceed;
+    }
+
+    virtual kstatus run_batch(std::vector<std::unique_ptr<ral::cache::CacheData>> batches, int batch_count) {
```
felipeblazing, line 69: why are you passing batch_count if you aren't using it?

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
```diff
@@ -85,7 +113,7 @@ class ComputeAggregateKernel : public kernel {
                 }
 
                 this->add_to_output_cache(std::move(output));
-                batch_count++;
+
```
felipeblazing, line 116: how are you tryacking the number of batches processed by the kernel without this?

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
```diff
@@ -131,13 +149,20 @@ class ComputeAggregateKernel : public kernel {
 private:
     std::vector<AggregateKind> aggregation_types;
     std::vector<int> group_column_indices;
+    std::vector<std::string> aggregation_input_expressions;
+    std::vector<std::string> aggregation_column_assigned_aliases;
```
felipeblazing, line 153: assigned_aliases? assigned by who for what? this name isn't clear to me perhaps it will be more clear if you ad a comment describing what this is for

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
```diff
@@ -131,13 +149,20 @@ class ComputeAggregateKernel : public kernel {
...
 };
 
 class DistributeAggregateKernel : public kernel {
 public:
 	DistributeAggregateKernel(std::size_t kernel_id, const std::string & queryString, std::shared_ptr<Context> context, std::shared_ptr<ral::cache::graph> query_graph)
 		: kernel{kernel_id, queryString, context, kernel_type::DistributeAggregateKernel} {
         this->query_graph = query_graph;
+        std::tie(this->group_column_indices, this->aggregation_input_expressions, this->aggregation_types,
```
felipeblazing, line 161: why did this change?

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
```diff
@@ -284,24 +304,62 @@ class MergeAggregateKernel : public kernel {
...
         bool ordered = false; // If we start using sort based aggregations this may need to change
-        BatchSequence input(this->input_cache(), this, ordered);
-        int batch_count=0;
+        BatchSequenceBypass input(this->input_cache(), this);
         try {
             while (input.wait_for_next()) {
-                auto batch = input.next();
-                // std::cout<<"MergeAggregateKernel batch "<<batch_count<<std::endl;
-                // ral::utilities::print_blazing_table_view_schema(batch->toBlazingTableView(), "MergeAggregateKernel_batch" + std::to_string(batch_count));
-                batch_count++;
```
felipeblazing, line 304: how are you keeping track of batch_count now?

engine/src/execution_graph/logic_controllers/BatchProcessing.h
```diff
@@ -307,7 +307,7 @@ class DataSourceSequence {
 	}
 
 	bool has_next() {
-		return (is_empty_data_source && batch_index < 1) || (is_gdf_parser && batch_index.load() < n_batches) || (cur_file_index < n_files);
+		return (is_empty_data_source && batch_index.load() < 1) || (is_gdf_parser && batch_index.load() < n_batches) || (cur_file_index < n_files);
```
felipeblazing, line 310: why did you add the load here instead of comparing to the atomic itself. this should not matter when comparing to a literal.

felipeblazing: https://stackoverflow.com/questions/18850752/must-i-call-atomic-load-store-explicitly

felipeblazing: or any of these actually

engine/src/execution_graph/logic_controllers/BatchProcessing.h
```diff
@@ -755,10 +821,27 @@ class Print : public kernel {
...
 		std::lock_guard<std::mutex> lg(print_lock);
-		BatchSequence input(this->input_cache(), this);
+		BatchSequenceBypass input(this->input_cache(), this);
+		int batch_count = 0;
 		while (input.wait_for_next() ) {
-			auto batch = input.next();
+			std::vector<std::unique_ptr<ral::cache::CacheData>> batches;
+			batches.push_back(input.next());
+			
+			status = run_batch(std::move(batches), batch_count);
```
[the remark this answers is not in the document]
williamBlazing: In its current implementation, yes. But that will be handled by what i am proposing here:
https://github.com/BlazingDB/blazingsql/issues/924

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
williamBlazing: See here https://github.com/BlazingDB/blazingsql/issues/924

williamBlazing: The short answer is that, correct, you dont want it to be a unique_ptr. This PR is meant to be one small step towards a broader goal. Working forward in smaller steps can help the bigger goal not be one super massive PR

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
williamBlazing: batch_count here is actually more like a place holder. The idea is that there will be more info passed along with the batches, with which you can setup priorities.

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
williamBlazing: This moved to line 55

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
williamBlazing: This is not new code, just moved. The asigned aliases are asigned by the user in the query. Think:
`select max(colA) as the_best_max from ....`
the_best_mas would be the assigned alias. And if none is assigned by the user, we define a column name based on the aggregation and column.

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
williamBlazing: This just moved this code from the run() function tot he constructor

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
williamBlazing: this run function only has one batch, so its hardcoded to 0 in line 325

williamBlazing: well... let me refrase that. It only has one execution batch. It taked in several batches and concatenates them. But the concept of batch here should be execution batch

engine/src/execution_graph/logic_controllers/BatchProcessing.h
williamBlazing: Using load or not does not matter. It does the same underneath, as your stack overflow post suggests. But using load explicitly indicates that its an atomic

williamBlazing:
too outdated

Issue #903, opened 2020-07-29
Run only some kernels at a time


williamBlazing:
In `graph::execute()` every `kernel::run()` function is launched simultaneously in a separate thread. This makes it so that its easy to run out of memory. 

We want to implement is limiter on the number of kernel threads that is configurable by config_options['MAX_KERNEL_RUN_THREADS'] = N 

Only up to N threads can be running simultaneously. As soon as a kenel run thread is complete another one, next on the list should be launched.

rommelDB:
Closed by #915

PR #915, opened 2020-07-31
[REVIEW] Adding max kernel num threads pool

rommelDB:
Implements #903

### DIFF
--- a/engine/src/CalciteInterpreter.cpp
+++ b/engine/src/CalciteInterpreter.cpp
@@ -106,10 +106,16 @@ std::vector<std::unique_ptr<ral::frame::BlazingTable>> execute_plan(std::vector<
 			// useful when the Algebra Relacional only contains: ScanTable (or BindableScan) and Limit
 			query_graph->check_for_simple_scan_with_limit_query();
 
+			size_t max_kernel_run_threads = 4; //default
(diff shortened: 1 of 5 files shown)


pyblazing/pyblazing/apiv2/context.py
```diff
@@ -1070,6 +1070,9 @@ def __init__(
             MEMORY_MONITOR_PERIOD : How often the memory monitor checks memory
                     consumption. The value is in milliseconds.
                     default: 50  (milliseconds)
+            MAX_KERNEL_RUN_THREADS : The number of threads available to run
+                    kernels simultaneously.
+                    default: 4
```
felipeblazing, line 1075: this number is too low it should be something more like 16

engine/src/CalciteInterpreter.cpp
```diff
@@ -106,10 +106,16 @@ std::vector<std::unique_ptr<ral::frame::BlazingTable>> execute_plan(std::vector<
 			// useful when the Algebra Relacional only contains: ScanTable (or BindableScan) and Limit
 			query_graph->check_for_simple_scan_with_limit_query();
 
+			size_t max_kernel_run_threads = 4; //default
```
felipeblazing, line 109: default should be higher. should not be sset here and in another place like i see at the bottom.

PR #931, opened 2020-08-08
[REVIEW] adding message validation tests

kharoc:


tests/BlazingSQLTest/Runner/runTest.py
```diff
@@ -991,6 +994,22 @@ def saveLogInFile(df):
     filepath = getFileName(dir_log)
     df.to_excel(filepath, index=False)
 
+def validate_messages(error_message):
```
williamBlazing, line 997: `validate_messages` should not be a standard validation for all queries. Each of these queries we are validating for an error message should be passed in the expected validation message. We will want to be able to test many many more types of error messages and we cant just have one catch all function. That is not something we can scale. You need to be able to pass in a specific message to run_query or have a different function similar to run_query but special for these types of tests.

kharoc:
rerun tests

kharoc:
rerun tests

williamBlazing:
rerun tests

kharoc:
rerun tests

PR #936, opened 2020-08-11
[WIP] Adding extern C for include files

### LINKED ISSUE #935: Extern C for include files
The only thing missing to make bindings to blazings C++ back end is make some C wrappers so that other languages that can bind to C but not C++

rommelDB:
Fixes #935

### DIFF
--- a/engine/src/error.hpp
+++ b/engine/src/error.hpp
@@ -33,13 +33,10 @@ struct BlazingMissingMetadataException : public std::exception
     std::string key;
 };
 
-/*enum _error
+enum _error
 {
   E_SUCCESS = 0,
-  E_EXCEPTION = -1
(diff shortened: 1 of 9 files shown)


engine/src/error.hpp
```diff
@@ -32,3 +32,14 @@ struct BlazingMissingMetadataException : public std::exception
...
 };
+
+/*enum _error
+{
+  E_SUCCESS = 0,
+  E_EXCEPTION = -1
+};
+
+typedef enum _error error_code_t;*/
+#define E_SUCCESS true
```
felipeblazing, line 43: make this an int please

rommelDB: Fixed.

PR #944, opened 2020-08-13
[PAUSED] Refactor message distribution

rommelDB:


engine/src/cython/engine.cpp
```diff
@@ -276,7 +276,7 @@ TableScanInfo getTableScanInfo(std::string logicalPlan){
 	return TableScanInfo{relational_algebra_steps, table_names, table_columns};
 }
 
-std::pair<std::unique_ptr<PartitionedResultSet>, error_code_t> runQuery_C(int32_t masterIndex,
+std::pair<std::shared_ptr<ral::cache::graph>, error_code_t> runGenerateGraph_C(int32_t masterIndex,
```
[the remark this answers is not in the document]
rommelDB: These changes are due to the C wrappers were merged to branch-0.15 before the update. Since there were changes in the UCX branch (specifically on the run_query function), some conflicts were raised. So, now those conflicts are fixed and the branch compiles ok again.

engine/src/distribution/MessageManager.cpp
```diff
@@ -0,0 +1,48 @@
...
+int MessageManager::get_total_partition_counts(std::map<std::string, int>& node_count) {
+    int total_count = node_count[node_id];
+    for (auto message : messages_to_wait_for){
+        auto meta_message = input_message_cache->pullCacheData(message);
+        total_count += std::stoi(static_cast<ral::cache::GPUCacheDataMetaData *>(meta_message.get())->getMetadata().get_values()[ral::cache::PARTITION_COUNT]);
+    }
+    return total_count;
+}
+
+void MessageManager::send_total_partition_counts(ral::cache::CacheMachine* graph_output,
```
felipeblazing, line 16: dont pass naked pointers

engine/src/distribution/MessageManager.cpp
```diff
@@ -0,0 +1,48 @@
...
+        std::string message_prefix,
+        std::string metadata_label,
+        std::string cache_id,
+        std::map<std::string, int>& node_count) {
```
felipeblazing, line 20: dont pass this is by reference this is bad practice. Return the values you need from the function

engine/src/distribution/MessageManager.cpp
```diff
@@ -0,0 +1,48 @@
+#include "MessageManager.h"
```
felipeblazing, line 1: Now instead of having the logic of  managing the waiting and sending of these messages in a common place it has been split up. This is not a sound pattern its too easy for things to get out of whack and its not clear.

engine/src/distribution/MessageManager.cpp
felipeblazing, line 16: change class to message_manager

engine/src/distribution/MessageManager.h
```diff
@@ -0,0 +1,47 @@
...
+#include <map>
+#include "execution_graph/logic_controllers/CacheMachine.h"
+#include <blazingdb/manager/Context.h>
+
+namespace ral {
+namespace distribution {
+
+using Context = blazingdb::manager::Context;
+
+class MessageManager {
```
felipeblazing, line 13: This  class seems to make the assumption that a message is only destined for one place which is incorrect.

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
```diff
@@ -136,11 +137,13 @@ class ComputeAggregateKernel : public kernel {
 class DistributeAggregateKernel : public kernel {
 public:
 	DistributeAggregateKernel(std::size_t kernel_id, const std::string & queryString, std::shared_ptr<Context> context, std::shared_ptr<ral::cache::graph> query_graph)
-		: kernel{kernel_id, queryString, context, kernel_type::DistributeAggregateKernel} {
-        this->query_graph = query_graph;
+		: kernel{kernel_id, queryString, context, kernel_type::DistributeAggregateKernel},
+		  message_manager{kernel_id, context, ral::communication::CommunicationData::getInstance().getSelfNode().id(), ral::communication::CommunicationData::getInstance().getSelfNode(),
```
felipeblazing, line 141: Do not instantiate any classes with some global function call. The more we do this the less we can write tests for things. Do this on the outside before calling the constructor.

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
```diff
@@ -237,6 +240,9 @@ class DistributeAggregateKernel : public kernel {
                         // }
                         // ral::distribution::distributeTablePartitions(this->context.get(), partitions_to_send);
 
+                        //assuming 1:1 a table for each node
```
felipeblazing, line 243: this coment is confusing

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
```diff
@@ -580,31 +587,7 @@ class JoinPartitionKernel : public kernel {
 			}
 		}
 
-		auto nodes = local_context->getAllNodes();
```
felipeblazing, line 583: leave this logic where it currently lies just take out the repetitive stuff that actually generates the metadata.

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
```diff
@@ -1094,6 +1069,9 @@ class JoinPartitionKernel : public kernel {
 	std::vector<cudf::size_type> left_column_indices, right_column_indices;
 	std::vector<cudf::data_type> join_column_common_types;
 	bool normalize_left, normalize_right;
+	ral::distribution::MessageManager message_manager_left;
```
felipeblazing, line 1072: it seems pretty silly to have  message_manager per kernel. Instead just make a few utility functions and leave the logic the same as i twas before.

engine/src/distribution/MessageManager.cpp
williamBlazing: Most of our class names are in camel case. We don't have one very specific way of naming I think class name should be in camel case and function name should be with underscores.

engine/src/distribution/MessageManager.cpp
williamBlazing: `node_count` should be a member of the class it's part of the state of the class. It should also be thread safe

engine/src/distribution/MessageManager.h
williamBlazing: Why do you say that I don't believe it makes that assumption at all

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
williamBlazing: Here it is instantiating the message manager for this particular kernel. It is part of the kernel. So it should be instantiated in the Constructor of the kernel

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
williamBlazing: One of the key ideas behind the message manager is that it maintains the state necessary for proper sending and receiving of messages such as the node_count

engine/src/distribution/MessageManager.cpp
felipeblazing: your comment seems to be addressing some other point. I am saying you cannot pass this map in here by reference.

engine/src/distribution/MessageManager.cpp
felipeblazing: We agreed to follow cudf conventions. The conventions cudf follows are lower case with _ for classes and functions

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.h
felipeblazing:
```
ral::communication::CommunicationData::getInstance().getSelfNode().id()

```
im referring to that

engine/src/distribution/MessageManager.cpp
williamBlazing: `node_count` should not be passed in at all

engine/src/distribution/MessageManager.cpp
williamBlazing: ugh... you are right. we we should just follow cudf. I'm just not very fond of having functions and classes have the same convention

engine/src/distribution/MessageManager.cpp
williamBlazing: can you explain what you mean

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
williamBlazing: One of the key ideas behind the message manager is that it maintains the state necessary for proper sending and receiving of messages such as the node_count, or PARTITION_COUNT. this communication Paradigm relies on certain usage patterns. To make it safer to use it's better to encapsulate the usage patterns in function calls

engine/src/distribution/MessageManager.cpp
felipeblazing: I mean the logic that does send messages and wait for those messages to count are very related and should be done in the smae logical place. Here the generation of the message and the actual waiting for its other parts to arrive is now seperated

engine/src/distribution/MessageManager.cpp
felipeblazing: I am. and it follows cudf

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
felipeblazing: The kernel should manage this as it did before.  Basically you are performring broadcast / gather. I would be ok with a broadcast function and one for gather that was a bit more clear but this  just seems more confusing

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
felipeblazing: What I dont get is that the logic is rather simple before. Send a list of messages, wait for their counterparts to arrive. The only thing that was really repetitive was the generate of the message metadata. I am saying this class feels a bit pointless

williamBlazing:
@rommelDB I talked with @felipeblazing and we came to an agreement on a small change in the design. Rather than having message manager being a separate class, should be an interface that inherits from the kernel class. So:

`class MessageManager` 
becomes:
`class distributing_kernel : public kernel`
And 
` class DistributeAggregateKernel : public kernel`
becomes:
` class DistributeAggregateKernel : public distributing_kernel`

This will make the implementation or the join kernel a little trickier. Since in the join kernel we want to have separate management of messages for the different tables.

engine/src/execution_graph/logic_controllers/taskflow/distributing_kernel.cpp
```diff
@@ -0,0 +1,96 @@
+#include "distributing_kernel.h"
+#include "utilities/CommonOperations.h"
+
+namespace ral {
+namespace cache {
+
+distributing_kernel::distributing_kernel(std::size_t kernel_id,
```
williamBlazing, line 7: @rommelDB you should be able to have the same function signature for the Constructor as the base kernel class

rommelDB: Fixed.

engine/src/distribution/MessageManager.cpp
rommelDB: Fixed.

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
```diff
@@ -523,6 +524,18 @@ class JoinPartitionKernel : public kernel {
 						partitioned.push_back(batch_view);
 					}
 				}
+
```
williamBlazing, line 527: I think here you are sending all the partitions to all the workers as you're supposed to. And I think the following block of code is no longer necessary because it seems that it is also sending all the data

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
```diff
@@ -472,18 +476,15 @@ class JoinPartitionKernel : public kernel {
 													right_join_types.cbegin(), right_join_types.cend());
 	}
 
-	static void partition_table(const std::string & kernel_id,
+	void partition_table(const std::string & kernel_id,
```
williamBlazing, line 479: This function is used for partitioning data for either left or right table. This means that depending on the table you are using the message tracker 0 or 1. And it also changes the destination table and the destination output cache. I don't think this is being done correctly please take another look

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
```diff
@@ -911,21 +879,19 @@ class JoinPartitionKernel : public kernel {
...
-						metadata.add_value(ral::cache::QUERY_ID_METADATA_LABEL, std::to_string(this->context->getContextToken()));
-						metadata.add_value(ral::cache::ADD_TO_SPECIFIC_CACHE_METADATA_LABEL, "true");
-						metadata.add_value(ral::cache::CACHE_ID_METADATA_LABEL, small_output_cache_name);
-						metadata.add_value(ral::cache::SENDER_WORKER_ID_METADATA_LABEL, self_node.id());
-						metadata.add_value(ral::cache::WORKER_IDS_METADATA_LABEL, worker_ids_metadata);
-						ral::cache::CacheMachine* output_cache = this->query_graph->get_output_message_cache();
-						output_cache->addCacheData(std::make_unique<ral::cache::GPUCacheDataMetaData>(small_table_batch->toBlazingTableView().clone(), metadata),"",true);
+						increment_node_count(self_node.id(), 0);
+
+						send_message(std::move(small_table_batch->toBlazingTableView().clone()),
```
williamBlazing, line 887: here we want to use a broadcast function. We are sending the same table to every node.

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.h
```diff
@@ -369,19 +353,23 @@ class PartitionKernel : public kernel {
...
-						output_cache->addCacheData(std::unique_ptr<ral::cache::GPUCacheData>(new ral::cache::GPUCacheDataMetaData(table_view.clone(), metadata)),"",true);
+						send_message(std::move(table_view.clone()),
+							"true", //specific_cache
+							"output_" + std::to_string(part_ids[i]), //cache_id
+							dest_node.id(), //target_id
+							"", //total_rows
+							"", //message_id_prefix
+							true); //always_add
+
+						increment_node_count(dest_node.id(), 0);
```
williamBlazing, line 364: `send_message` should be the one that does `increment_node_count`

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.h
```diff
@@ -369,19 +353,23 @@ class PartitionKernel : public kernel {
...
 					}
 
 					for (auto i = 0; i < partitions.size(); i++) {
 						auto & partition = partitions[i];
```
williamBlazing, line 368: we should make `send_message` know what to do with sending to `self_node` that way we dont need this sort of logic for messages not actually being sent anywhere

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
rommelDB: Yes, that code was actually replicated. I removed it.

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
rommelDB: You're right, it was inconsistent. I changed the logic.

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.h
rommelDB: Fixed!

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.h
rommelDB: Done. Now the increment is performed inside the function send_message.

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.h
rommelDB: This change is not as simple as I thought, I mean, this pattern could apparently be replaced by a scatter but a bit more general.

williamBlazing:
This work is to be completed in this PR https://github.com/BlazingDB/blazingsql/pull/1132

PR #950, opened 2020-08-18
[REVIEW] introducing config param for max orderby samples and fixing oversampling

williamBlazing:


williamBlazing:
By oversampling I mean that we were collecting more samples and we're actually being used hey

PR #957, opened 2020-08-19
Dummy PR

williamBlazing:


williamBlazing:
> This looks good are there any tests for this functionality in the unit tests?

honestly i am not sure

engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -110,10 +117,96 @@ struct cast_to_timestamp_ns {
...
+struct cast_to_date {
+	template <typename Element, std::enable_if_t<!cudf::is_timestamp<Element>() && !(cudf::is_numeric<Element>() && !cudf::is_floating_point<Element>())> * = nullptr>
+	CUDA_DEVICE_CALLABLE cudf::timestamp_D operator()(cudf::type_id type_id, int64_t val) {
+		assert(false);
+		return cudf::timestamp_D{};
+	}
+
+	template <typename Timestamp, std::enable_if_t<cudf::is_timestamp<Timestamp>() || (cudf::is_numeric<Timestamp>() && !cudf::is_floating_point<Timestamp>())> * = nullptr>
+	CUDA_DEVICE_CALLABLE cudf::timestamp_D operator()(cudf::type_id type_id, int64_t val) {
+		cudf::timestamp_D computed;
```
jeanp413, line 133: With `!cudf::is_floating_point<Element>()` added, casting from float or float64 (which implicitly downcasts to an integer) to date or timestamp will no longer work. Is that expected?
Also there is no point in using type_dispatcher if you are passing the `type_id` and doing a switch, just call the function directly

williamBlazing: Good point Thank you for pointing that out. I had originally used the type dispatcher trying to fix an issue that a double was causing. But now I realize that I don't need it

PR #961, opened 2020-08-20
[REVIEW] Add list_tables() and describe_table() functions

### LINKED ISSUE #954: Implement bc.list_tables() and bc.describe_table(table_name)
Look into what calcite returns for DDL... 
update apache calcite does not support DDL out of the box. See here https://calcite.apache.org/docs/adapter.html#server

Christian8491:
This PR add two functions:
* list_tables() : returns a list with all the created tables
* describe_table(table_name) : returns the column names and column types as a dictionary.
And a doc about these functions.
Closes #954

### DIFF
--- a/pyblazing/pyblazing/apiv2/context.py
+++ b/pyblazing/pyblazing/apiv2/context.py
@@ -1905,7 +1905,7 @@ class BlazingContext(object):
     def describe_table(self, table_name):
         """
         Returns a dictionary with the names of all the columns and their types
-        for the specified table. A ValueError is thrown if thr table is not found.
+        for the specified table. A ValueError is thrown if the table is not found.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


pyblazing/pyblazing/apiv2/context.py
```diff
@@ -1876,6 +1876,59 @@ def drop_table(self, table_name):
...
+        Example
+        --------
+
+        >>> from blazingsql import BlazingContext
+        >>> bc = BlazingContext()
+        >>> bc.create_table('nation', "nation/*.parquet")
+        >>> info_table = bc.describe_table("nation")
+        >>> print(info_table)
+                  {'n_nationkey': 'int32', 'n_name': 'object',
+                   'n_regionkey': 'int32', 'n_comment': 'object'}
```
williamBlazing, line 1916: is the type shown for a string column `object`? i understand why, but i think we want it to display `str`

williamBlazing: have a funtion which maps the data types to user friendly strings to display to the user

Christian8491: Updating:   `object`   to `str`

PR #963, opened 2020-08-20
[REVIEW] Set log_level when using LOGGING_LEVEL param

### LINKED ISSUE #953: add configurable log level
right now logging is impacting run time and theres no way to prevent that. We should wrap our logging statements in some kind of macro that lets us turn it on and off depending on levels set. Right now we only have something like that for flushing

Christian8491:
This PR uses the param `LOGGING_LEVEL` for set the level of logs we want. So the new name for the flush_level is `LOGGING_FLUSH_LEVEL`

Closes #953

### DIFF
--- a/engine/src/cython/initialize.cpp
+++ b/engine/src/cython/initialize.cpp
@@ -61,24 +61,29 @@ std::string get_ip(const std::string & iface_name = "eth0") {
 	return the_ip;
 }
 
-void select_flush_level(std::string flush_level) {
-	if (flush_level == "critical") {
-		spdlog::flush_on(spdlog::level::critical);
+
+
+auto log_level_str_to_enum(std::string level) {
+	if (level == "critical") {
+		return spdlog::level::critical;
 	}
-	else if (flush_level == "err") {
-		spdlog::flush_on(spdlog::level::err);
+	else if (level == "err") {
+		return spdlog::level::err;
 	}
-	else if (flush_level == "info") {
-		spdlog::flush_on(spdlog::level::info);
+	else if (level == "info") {
+		return spdlog::level::info;
 	}
-	else if (flush_level == "debug") {
-		spdlog::flush_on(spdlog::level::debug);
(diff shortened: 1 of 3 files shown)


engine/src/cython/initialize.cpp
```diff
@@ -73,17 +78,48 @@ void create_logger(std::string fileName, std::string loggingName, int ralId, std
...
+	
+	// By default, will contains all sort of logs (more logs detail)
+	std::string logger_level_wanted = "trace";
+	auto log_level_it = config_options.find("LOGGING_LEVEL");
+	if (log_level_it != config_options.end()){
+		logger_level_wanted = config_options["LOGGING_LEVEL"];
+	}
+
+	// level of logs
+	if (logger_level_wanted == "off") {
```
williamBlazing, line 94: This mapping of string to spdlog enum is now happening twice can you please turn this into a function we can reuse.

engine/src/cython/initialize.cpp
```diff
@@ -62,7 +62,12 @@ std::string get_ip(const std::string & iface_name = "eth0") {
 }
 
 // simple_log: true (no timestamp or log level)
-void create_logger(std::string fileName, std::string loggingName, int ralId, std::string flush_level, bool simple_log=true){
+void create_logger(std::string fileName,
+	std::string loggingName,
+	int ralId, std::string flush_level,
+	std::map<std::string, std::string> config_options,
```
williamBlazing, line 68: Here please pass in the logging level and not the config_options

engine/src/cython/initialize.cpp
Christian8491: Currently `logger_level` is just once but the `flush_level` is mapping twice. So I am updating this `flush_level` in a separate function then we can reuse it.

engine/src/cython/initialize.cpp
Christian8491: ok. Updated.

PR #965, opened 2020-08-24
[REVIEW] Handle exceptions from pool_threads

Christian8491:
Throwing manually an exception on the  `run()` method from `BindableTableScan`.

```
BlazingContext ready
[16:05:36.925] [error] 256896318|1|1|In BindableTableScan kernel batch for BindableTableScan(table=[[main, nation]], filters=[[<($1, 3)]], projects=[[0, 2]], aliases=[[n_nationkey, $f1]]). What: Throwing a runtime exception inside BindableTableScan|||||
[16:05:36.926] [error] 256896318|0|0|In execute_plan. What: An exception occurred when run() method was called|||||
An exception occurred when run() method was called[16:05:36.926] [error] 256896318|0|0|In runQuery. What: An exception occurred when run() method was called|||||

>>>>>>>>  [RunQuery Error] An exception occurred when run() method was called
Empty DataFrame
Columns: []
Index: []
Series([], dtype: float64)
```

### DIFF
--- a/engine/src/distribution/primitives.cpp
+++ b/engine/src/distribution/primitives.cpp
@@ -205,7 +205,8 @@ void distributeTablePartitions(Context * context, std::vector<NodeColumnView> & 
 	if (it != config_options.end()){
 		max_message_threads = std::stoi(config_options["MAX_SEND_MESSAGE_THREADS"]);
 	}
-	ctpl::thread_pool<BlazingThread> threads(max_message_threads);
+	ctpl::thread_pool<BlazingThread> pool(max_message_threads);
+	std::vector<std::future<void>> futures;
 	for (auto i = 0; i < partitions.size(); i++){
 		auto & nodeColumn = partitions[i];
 		if(nodeColumn.first == self_node) {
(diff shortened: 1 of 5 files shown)


engine/src/distribution/primitives.cpp
```diff
@@ -263,10 +269,16 @@ void distributePartitions(Context * context, std::vector<NodeColumnView> & parti
 		}
 		BlazingTableView columns = nodeColumn.second;
 		auto destination_node = nodeColumn.first;
-		threads.push(([message_id, context_token, self_node, destination_node, columns](int thread_id) mutable {
+		auto send_function = threads.push(([message_id, context_token, self_node, destination_node, columns](int thread_id) mutable {
```
felipeblazing, line 272: this wont work because basically every kernel jsut became single threaded with this use case.

engine/src/distribution/primitives.cpp
```diff
@@ -263,10 +269,16 @@ void distributePartitions(Context * context, std::vector<NodeColumnView> & parti
...
 			auto message = Factory::createColumnDataMessage(message_id, context_token, self_node, columns);
 			Client::send(destination_node, *message);
 		}));
+
+		try {
+			send_function.get();
```
felipeblazing, line 278: blocks execution when it should be able to schedule as many of these as it needs. this needs to be caught by some global context that can then STOP the whole graph from executing on EVERY node

felipeblazing: To be clear we dont have to stop execution on every node for now but if we are using a pool we MUST have some lace to store these futures to make sure they all have get run on them before the query executes but they cant prevent the kernel from scheduling multiple simultaneously

PR #972, opened 2020-08-27
[WIP] removed persist for dask dataframe table creation

williamBlazing:


williamBlazing:
This did not really work at scale

PR #975, opened 2020-08-31
[WIP] fixed issue due to cudf orc api change

williamBlazing:
Fixed minor issue due to cudf orc api. Also changed when cmake is called for cudf. It is now only run with `./build.sh update`.


williamBlazing:
rerun tests

PR #977, opened 2020-08-31
[REVIEW]  Improve allocator options and logging

Christian8491:
This PR adds back the pool allocators. It also adds the maximum_pool_size parameter and the enable_logging option, which enables the allocators to create their logs. 
This PR also adds logging to the allocator initialization. In order to add the logging to the initialization, it was necessary to place the allocator initialization to the initializeBlazing function. This also helps reduce the number of cython functions to maintain. This will cause some conflicts when merging the UCX branch, but they are tractable merge conflicts.

This PR fixes: #1050 and #1044


engine/src/cython/initialize.cpp
```diff
@@ -274,27 +292,15 @@ void finalize() {
 	exit(0);
 }
 
-void blazingSetAllocator(
```
felipeblazing, line 277: where is this api now?

engine/bsql_engine/io/io.pyx
```diff
@@ -168,15 +168,14 @@ cpdef pair[bool, string] registerFileSystemCaller(fs, root, authority):
     if fs['type'] == 'local':
         return cio.registerFileSystemLocal( str.encode( root), str.encode(authority))
 
-cpdef initializeCaller(int ralId, int gpuId, string network_iface_name, string ralHost, int ralCommunicationPort, bool singleNode, map[string,string] config_options):
-    initializePython( ralId,  gpuId, network_iface_name,  ralHost,  ralCommunicationPort, singleNode, config_options)
+cpdef initializeCaller(int ralId, int gpuId, string network_iface_name, string ralHost, int ralCommunicationPort, bool singleNode, map[string,string] config_options, string allocation_mode, size_t initial_pool_size):
+    initializePython(ralId, gpuId, network_iface_name, ralHost, ralCommunicationPort, singleNode, config_options, allocation_mode, initial_pool_size)
```
felipeblazing, line 172: how do they use the existing allocator when they dont want to make a new one?

williamBlazing: This is handled the same way as before, if you set the allocator to `existing` it will grab the current one and wrap it

engine/src/cython/initialize.cpp
williamBlazing: It is now in the singular initialize API

williamBlazing:
This PR also resolves #1044

PR #997, opened 2020-09-03
[REVIEW] Add capacity to set the transport memory

Christian8491:


Christian8491:
rerun tests

williamBlazing:
rerun tests

engine/src/cython/initialize.cpp
```diff
@@ -144,11 +144,15 @@ void initialize(int ralId,
...
+	auto it_pinned = config_options.find("TRANSPORT_BUFFER_BYTE_SIZE");
+	if (it_pinned != config_options.end()){
+		sizeBuffers = std::stoi(config_options["TRANSPORT_BUFFER_BYTE_SIZE"]);
+	}
 
-	size_t total_gpu_mem_size = ral::config::gpuTotalMemory();
-	assert(total_gpu_mem_size > 0);
 	auto nthread = 4;
-	blazingdb::transport::io::setPinnedBufferProvider(0.1 * total_gpu_mem_size, nthread);
+	blazingdb::transport::io::setPinnedBufferProvider(sizeBuffers, nthread);
```
felipeblazing, line 155: since we are now fixing the size of the buffers can we vary the amount it allocates to reflect gpu size by setting nthreads accordinglly? it feels like a v100 with 32GB and a 1050 with 2GB want a different amount here

felipeblazing (changes requested):

engine/src/cython/initialize.cpp
williamBlazing: I think the amount in nthreads that it actually needs is not going to be dependent on the GPU size. The number of buffers that it'll actually use will depend on the data that it's sending and to how many nodes its sending to and the number of transport threads (MAX_SEND_MESSAGE_THREADS). 
I do think that hardcoding nthreads is not necessarily a good idea. How about we set it to be the same as the value of: MAX_SEND_MESSAGE_THREADS

williamBlazing: And also `nthread` is actually the number of buffers that it starts off with, it has nothing to do with threads. The number of buffers will dynamically grow as a necessary

PR #1007, opened 2020-09-09
[REVIEW] fixing arrow compilation

Christian8491:
This PR also  fix issues related to new parquet API in cudf.

### DIFF
--- a/conda/recipes/blazingsql/meta.yaml
+++ b/conda/recipes/blazingsql/meta.yaml
@@ -28,6 +28,7 @@ requirements:
         - python
         - pip
         - setuptools
+                - spdlog=1.7.0
         - cython >=0.29,<0.30
         - openjdk >=8.0
         - maven
@@ -43,6 +44,7 @@ requirements:
         - python
         - pip
         - setuptools
+                - spdlog=1.7.0
         - cython >=0.29,<0.30
         - openjdk >=8.0
         - maven
--- a/conda/recipes/blazingsql/meta.yaml
+++ b/conda/recipes/blazingsql/meta.yaml
@@ -28,7 +28,7 @@ requirements:
         - python
         - pip
         - setuptools
-                - spdlog=1.7.0
+        - spdlog >=1.7.0,<1.8.0
         - cython >=0.29,<0.30
         - openjdk >=8.0
         - maven
@@ -44,7 +44,7 @@ requirements:
         - python
         - pip
         - setuptools
-                - spdlog=1.7.0
+        - spdlog >=1.7.0,<1.8.0
         - cython >=0.29,<0.30
         - openjdk >=8.0
         - maven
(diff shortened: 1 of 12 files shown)


io/src/FileSystem/private/GoogleCloudStorageReadableFile.cpp
```diff
@@ -178,7 +178,7 @@ return arrow::Status::OK();
 		//		results.GetResult().GetBody().read((char*)(out->get()->mutable_data()), bytesRead);
 
 		std::shared_ptr<arrow::ResizableBuffer> buffer;
-		AllocateResizableBuffer(arrow::default_memory_pool(), nbytes, &buffer);
+		AllocateResizableBuffer(nbytes, arrow::default_memory_pool());
```
aucahuasi, line 181: Should be something like this:
`buffer = AllocateResizableBuffer(nbytes, arrow::default_memory_pool());`
We need to make sure we are allocating the buffer

Christian8491: Sure. thanks for the advice.

conda/recipes/blazingsql/meta.yaml
```diff
@@ -28,6 +28,7 @@ requirements:
         - python
         - pip
         - setuptools
+        - spdlog >=1.7.0,<1.8.0
```
aucahuasi, line 31: Why we need to declare spdlog on here? Is already declared in the rmm package
https://github.com/rapidsai/rmm/blob/branch-0.16/conda/recipes/librmm/meta.yaml#L31
We should avoid to include here transitive dependencies cc @mario21ic

io/src/FileSystem/private/GoogleCloudStorageReadableFile.cpp
```diff
@@ -177,18 +178,18 @@ return arrow::Status::OK();
 		//}
 		//		results.GetResult().GetBody().read((char*)(out->get()->mutable_data()), bytesRead);
 
-		std::shared_ptr<arrow::ResizableBuffer> buffer;
-		AllocateResizableBuffer(arrow::default_memory_pool(), nbytes, &buffer);
+		arrow::Result<std::unique_ptr<arrow::ResizableBuffer>> buffer = AllocateResizableBuffer(nbytes, arrow::default_memory_pool());
 
-		results.read((char *) (buffer->mutable_data()), bytesRead);
+		results.read((char *) (buffer.ValueOrDie()->mutable_data()), bytesRead);
```
aucahuasi, line 183: How we handle the "die" part? For example that arrow API says
`buffer.ValueOrDie()`
I think in most cases we will get the buffer but there will be be some cases where we get a "die" object/buffer ... how do we handle that? maybe we need to raise with low exception here cc @felipeblazing

conda/recipes/blazingsql/meta.yaml
williamBlazing: without it, its picking the latest spdlog 1.8.0 which we dont want.Not sure why the rmm one is not being used

io/src/FileSystem/private/GoogleCloudStorageReadableFile.cpp
williamBlazing: The exception handling of all of this code is in desperate need of cleanup. It has been outdated for a long time. I think that needs to be addressed by a different PR

PR #1011, opened 2020-09-14
[WIP] Fix bc.partition

williamBlazing:


williamBlazing:
rerun-tests

williamBlazing:
rerun-tests

williamBlazing:
One reason this is not working correctly is that when using a dask_DataFrame partitioned by bc.partition, dask is assuming that every dask partition contains a unique set of keys of the partitioned column(s). But in the current implementation here, we are assuming that every bsql worker containers a unique set of keys in the partitioned column(s) and each worker can have a multiple partitions. So to fix this we have two options:
1. quick option is to concat all partitions before outputting. This has the downside of perhaps making the partitions too big. This would be really fast to implement but in the long run we may need to improve.
2. We could break up the partitions into multiple output partitions the way that order by merge does. This would involve a fair amount more logic. The total amount of partitions could be an extra parameter provided by the user. 

I would recommend implementing the quick version for now and then the improved version after the UCX communications refactor is merged in.

williamBlazing:
Now that new comms has been merged in, this PR can be continued and completed.

To summarize what this PR should do:
bc.partition is meant to partition data based on the hashes of the columns passed into the `partition_by` parameter. This would make it so that every partition will have a unique set of keys that would not be present in any other partition.

This PR will introduce two new kernels. 
- **SingleTableHashPartitionKernel**: this kernel is mostly implemented. It will need to be updated to work with new comms. Additionally we need to change the logic so that the data is being partitioned into `num_partitions_per_worker` X `num_workers` = `total_num_partitions` . This will mean that there are multiple partitions per worker and the way to handle that and send those messages will be similar to how `PartitionKernel` from `BatchOrderByProcessing` does it.

- **ConcatenatingPartitionsKernel**: This kernel will be similar to `MergeStreamKernel` from `BatchOrderByProcessing` except that instead of using `cudf::merge` it will concatenate


this is what needs to be done to complete this PR:

1. bc.partition needs a new parameter which is num_partitions_per_worker which needs to be 1 or greater
2. The CacheMachines between `SingleTableHashPartitionKernel` and `ConcatenatingPartitionsKernel` will need to be multiple partitions defined by `num_partitions_per_worker` similarly to how the CacheMachines between `PartitionKernel` and `MergeStreamKernel` are defined by `MAX_NUM_ORDER_BY_PARTITIONS_PER_NODE`
3. Need to update `SingleTableHashPartitionKernel` to use new comms and to shuffle to multiple partitions per worker
4. Need to create a new kernel `ConcatenatingPartitionsKernel`. To see everything that needs to be done for a new kernel, see the rest of this PR to see what was done to intruduce the new `SingleTableHashPartitionKernel` kernel

PR #1017, opened 2020-09-17
[REVIEW] fixed issue where fixed size chars were causing the engine to crash, …

felipeblazing:
if the user tried to output a char literal it could be cast like this by calcite

cast():CHAR(12)

The engine did not support this operation


felipeblazing:
rerun tests

PR #1022, opened 2020-09-18
[WIP] Fix pinned buffer pool

williamBlazing:


engine/src/cython/initialize.cpp
```diff
@@ -173,7 +173,7 @@ void initialize(int ralId,
 	std::string env_cuda_device_str = env_cuda_device == nullptr ? "" : std::string(env_cuda_device);
 	initLogMsg = initLogMsg + "CUDA_VISIBLE_DEVICES is set to: " + env_cuda_device_str + ", ";
 	
-	size_t buffers_size = 78643200;  // 75 MBs        0.1 * free_gpu_mem_size;
+	size_t buffers_size = 78643200 / 8;  // 75 MBs / 8       0.1 * free_gpu_mem_size;
```
williamBlazing, line 176: We are going to change this value, we need to change it everywhere in the code that we refer to the defaults of TRANSPORT_BUFFER_BYTE_SIZE

comms/src/blazingdb/transport/io/reader_writer.cpp
```diff
@@ -29,42 +29,58 @@ namespace io {
 // numBuffers should be equal to number of threads
 PinnedBufferProvider::PinnedBufferProvider(std::size_t sizeBuffers,
                                            std::size_t numBuffers) {
+  this->bufferSize = sizeBuffers;
   this->buffer_counter = numBuffers;
-  for (int bufferIndex = 0; bufferIndex < numBuffers; bufferIndex++) {
+  this->allocations = std::vector<char *>(1);
+  cudaError_t err = cudaMallocHost((void **)&allocations[0], numBuffers * sizeBuffers * 5);
```
williamBlazing, line 35: If we define `numBuffers` we should use it, not just arbitrarily multiply it by 5. The number coming in is right now hardcoded to 20. We can hardcode that to 100, if we want, but we should take out the ` * 5` here

comms/src/blazingdb/transport/io/reader_writer.cpp
```diff
@@ -29,42 +29,58 @@ namespace io {
...
+  if (err != cudaSuccess) {
+    throw std::exception();
+  }
+  for (int bufferIndex = 0; bufferIndex < numBuffers * 5; bufferIndex++) {
```
williamBlazing, line 39: take out the ` * 5 ` here too

comms/src/blazingdb/transport/io/reader_writer.cpp
```diff
@@ -29,42 +29,58 @@ namespace io {
...
-    this->bufferSize = sizeBuffers;
-    cudaError_t err = cudaMallocHost((void **)&buffer->data, sizeBuffers);
-    if (err != cudaSuccess) {
-      throw std::exception();
-    }
+    buffer->size = this->bufferSize;
+    buffer->data = allocations[0] + bufferIndex * this->bufferSize;
     this->buffers.push(buffer);
   }
+  this->buffer_counter =  numBuffers * 5;
```
williamBlazing, line 45: take out the ` * 5 ` here too

comms/src/blazingdb/transport/io/reader_writer.cpp
```diff
@@ -29,42 +29,58 @@ namespace io {
...
 void PinnedBufferProvider::grow() {
-  this->buffer_counter++;
+  std::unique_lock<std::mutex> lock(inUseMutex);
+    
+
   PinnedBuffer *buffer = new PinnedBuffer();
   buffer->size = this->bufferSize;
-  cudaError_t err = cudaMallocHost((void **)&buffer->data, this->bufferSize);
+  allocations.resize(allocations.size() + 1);
+  cudaError_t err = cudaMallocHost((void **)&allocations[allocations.size() -1], this->bufferSize * 100);
```
williamBlazing, line 76: I would suggest growing the buffer pool by `numBuffers/2`  and have `numBuffers` be stored as a member of PinnedBufferProvider. That way we dont have another random hardcoded number. 
I suggest `numBuffers/2` because that is a common pattern in memory pools to grow by 50% of the initial amount.

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -394,6 +394,8 @@ def modifyAlgebraForDataframesWithOnlyWantedColumns(
 
 
 def get_uri_values(files, partitions, base_folder):
+    print("partitions: " + str(partitions))
```
williamBlazing, line 397: remove these prints

engine/src/cython/initialize.cpp
felipeblazing: sure in all honesty we need to set the default to a smaller buffer given the results of our experiments on allocation size

comms/src/blazingdb/transport/io/reader_writer.cpp
felipeblazing: agreed

comms/src/blazingdb/transport/io/reader_writer.cpp
felipeblazing: in all honesty i am kind of opposed to merging this in until we add something that will alos shrink the pool

pyblazing/pyblazing/apiv2/context.py
felipeblazing: never

felipeblazing: :)

PR #1028, opened 2020-09-21
[WIP] match dtypes after create table with multiple files

### LINKED ISSUE #1012: [BUG]  Fix bug where column types are ignored after create_table
https://github.com/BlazingDB/blazingsql/blob/branch-0.16/engine/src/cython/engine.cpp#L41

That line is quite clearly wrong. It assumes we only set the types for a table if those type swere specified in the kwargs but after you create a table the types are set in stone and MUST be for dask. Right now its inferring it for every file so that if two files types get inferred differently the types dont match up.

I literally cant run a select * on two files because of this

Christian8491:
This PR ensure matching dtype of columns after create table with multiple files that can have different dtype.
It closes #1012


Christian8491:
rerun tests

Christian8491:
rerun tests

williamBlazing:
rerun-tests

ajschmidt8:
rerun tests

PR #1030, opened 2020-09-22
[WIP] Avoid read _metadata files

### LINKED ISSUE #1024: [BUG] whenever _metadata exists in a folder where we are gathering parquet files blazing GLORIOUSLY ERUPTS into one massive segfault
i believe the title suffices but to be clear.

If you do something like 

```
bc.create_table("t1","/some/dir/with/parquet/files")
```

If that dir has a _metadata file in it we just crash crash crash trying to read it as a parquet file. Which really begs the question. Why DIDNT they make metadata a parquet file istelf. Unless it is.

Note: These kinds of whimsical poorly defined issues are a symptom of the weekend.

Christian8491:
This PR avoids read those _metadata files, normally in parquet directories.
It closes  #1024


Christian8491:
rerun tests

williamBlazing:
rerun-tests

PR #1020, opened 2020-09-18
[REVIEW] Updating rmm header, fixing e2e tests with deprecated drop_column function and fixing ORC args changes

rommelDB:
- Fix for building issues after updating the last rmm.
- Fix for columnBasisTest due to drop_column() is now deprecated.
- Fix for Orc arg changes.

PR #1035, opened 2020-09-23
[WIP] Fix CMake for RMM going header only

kkraus14:
Since RMM went header only we should no longer try to link against `rmm` in CMake (until CMake is revamped to support an `rmm` interface target).


williamBlazing:
rerun tests

williamBlazing:
rerun-tests

williamBlazing:
These changes were all placed in PR #1020

PR #1040, opened 2020-09-25
[REVIEW] Update conda recipe, remove cxx11 abi

kkraus14:
- Updates CMake
  - Removes CXX11 ABI handling (libcudf has removed this)
  - Remove linking to `cuda` as I believe driver API isn't used directly anywhere (someone please let me know if this is incorrect)
- Updates conda recipe
  - Skips building tests as part of conda-build
  - Removes `host` section for now since everything was duplicated between `build` and `host` and it caused some nasty workarounds (aws cpp handling)
  - Moves dependencies as possible to runtime only instead of build time
  - Uses `pin_compatible` jinja templating to guarantee compatibility between build-time and run-time dependencies.
  - Makes sure import test uses the proper cuda version

Just a note: lumping everything into a single conda package like this makes things quite difficult to manage from the conda build side. I.E. we should really have a `build` section for build tools like `cmake` / `ninja`, but doing so causes Python to be installed into the build prefix which messes up the Python parts of the build.


rommelDB:
Hi @kkraus14 , I'm working on a PR that is something complementary to this one. There https://github.com/BlazingDB/blazingsql/pull/1055 I'm removing the dependency on cudf that we had when some tests utilities were not public. So now, I'm wondering if there is any optimization that can be done in the compilation of the engine to remove the dependency on gtests in addition to the changes you made mainly when we building the engine without the tests.

conda/recipes/blazingsql/meta.yaml
```diff
@@ -20,54 +20,37 @@ source:
...
-        - maven
         - ninja
-        - dask-cudf {{ minor_version }}.*
-        - dask-cuda {{ minor_version }}.*
-        - cudatoolkit {{ cuda_version }}.*
-    host:
         - gtest
-        - gmock
+        - aws-sdk-cpp
         - google-cloud-cpp 1.16.0
```
kkraus14, line 27: This just finished migrationg to 1.18.0 in conda-forge so we may want to upgrade this as well.

kkraus14, line 27:
```suggestion
        - google-cloud-cpp
```

conda/recipes/blazingsql/meta.yaml
```diff
@@ -20,54 +20,37 @@ source:
...
         - setuptools
         - spdlog >=1.7.0,<1.8.0
         - cython >=0.29,<0.30
         - openjdk >=8.0
         - maven
-        - ninja
-        - dask-cudf {{ minor_version }}.*
-        - dask-cuda {{ minor_version }}.*
-        - cudatoolkit {{ cuda_version }}.*
+        - cudf {{ minor_version }}
```
kkraus14, line 35:
```suggestion
        - cudf {{ minor_version }}
        - dlpack
```

conda/recipes/blazingsql/meta.yaml
```diff
@@ -20,54 +20,39 @@ source:
...
-        - dask-cuda {{ minor_version }}.*
-        - cudatoolkit {{ cuda_version }}.*
-    host:
         - gtest
-        - gmock
-        - google-cloud-cpp 1.16.0
-        - cmake
+        - aws-sdk-cpp
+        - google-cloud-cpp 1.16.0.*
+        #- nlohmann_json
```
kkraus14, line 28: This was from trying to upgrade to 1.18.0 which has a bug of not including this automatically, see: https://github.com/conda-forge/google-cloud-cpp-feedstock/issues/28

conda/recipes/blazingsql/meta.yaml
```diff
@@ -20,54 +20,38 @@ source:
...
         - setuptools
         - spdlog >=1.7.0,<1.8.0
         - cython >=0.29,<0.30
         - openjdk >=8.0
         - maven
-        - ninja
-        - dask-cudf {{ minor_version }}.*
-        - dask-cuda {{ minor_version }}.*
+        - cudf {{ minor_version }}.*
+        - dlpack
```
kkraus14, line 36: We can remove this after https://github.com/rapidsai/cudf/pull/6445 is merged

PR #1046, opened 2020-09-29
[REVIEW] make possible to read sys env to setup blazingconf

### LINKED ISSUE #979: UX BlazingConfig hasmap should always read from context vars first
So the user have more flexibility to manage the conf

williamBlazing:
Fixes #979


Christian8491:
rerun tests

PR #1058, opened 2020-10-02
[WIP] Fix avg_bytes_per_row value when node 0 has no rows

Christian8491:


felipeblazing (changes requested):

Christian8491:
rerun-tests

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.h
```diff
@@ -118,6 +118,20 @@ class SortAndSampleKernel : public kernel {
 		return true;
 	}
 
+	// If Node 0 have no rows we want to stimate the `avg_bytes_per_row using its dtypes
+	size_t estimate_avg_bytes_per_row_using_dtypes(std::vector<cudf::data_type> dtypes) {
+		size_t avg_bytes_per_row = 0;
+		for (cudf::size_type i = 0; i < dtypes.size(); ++i) {
+			if (dtypes[i].id() == cudf::type_id::STRING) {
+				// TODO: we are assuming an average value for STRING columns
+				avg_bytes_per_row += 250;
```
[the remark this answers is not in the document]
Christian8491: updated, using the average of the samples to get a best `avg_bytes_per_row`

engine/src/operators/OrderBy.cpp
```diff
@@ -214,7 +214,34 @@ std::unique_ptr<ral::frame::BlazingTable> generate_partition_plan(const std::vec
 	}
 
 	int num_nodes = context->getTotalNodes();
-	cudf::size_type total_num_partitions = (double)table_num_rows*(double)avg_bytes_per_row/(double)num_bytes_per_order_by_partition;
+
+	// This mean master node does not have data, so we want to recompute the `avg_bytes_per_row` value
+	// using the average of all the samples that contains data
+	if (avg_bytes_per_row == 1 && samples.size() > 1 && table_num_rows > 0) {
```
williamBlazing, line 220: I understand what we are doing here, but this means we are now sending samples that have all the columns, but we ONLY need the columns, in case we need to estimate avg_bytes_per_row.

I think what we want to do instead is only send the columns we need, but add as extra metadata the avg_bytes_per_row for every sample. This will be able to be done more easily when we have the new communications.

I think we should postpone this PR until we have the new communications and implement it in such a way so that we are not sending those extra columns with the samples

williamBlazing (changes requested):
I propose we pause this PR until after the new communications is implemented. See my comments in the code

engine/src/operators/OrderBy.cpp
williamBlazing: @Christian8491 now the the new communications has been merged in, could you modify this PR so that when you send the samples you also send an extra metadata that is AVG_BYTES_PER_ROW. This would be a new metadata tag, but it should be fairly straight forward to implement. Please let me know if you need help with this

Christian8491:
This PR should be closed, as #1154 is in coming

Christian8491:
Closing this PR, as #1154 was merged.

PR #1060, opened 2020-10-05
[PAUSED] Replacing calls to the scheduler_info by self.nodes

### LINKED ISSUE #1026: [BUG] if a worker crashes ral_id gets out of synch
we iterator through workers like so
for worker in list(self.dask_client.scheduler_info()["workers"]):

and increment ral_id with a variable

instead we should store a map so we can maintain ral_id even if a worker dies

rommelDB:
Fixes #1026


felipeblazing (changes requested):
This doens't address the isue 1026 at all. Remember the worker can change names when it restarts. This should not be merged in

felipeblazing (changes requested):
No where are you verifying that the node shavent changed and are still in good order.

what value does this provide?

GPUtester:

PR #1064, opened 2020-10-06
[REVIEW] Fix for creating tables from distributed local parquet files

### LINKED ISSUE #1062: [BUG] Parquet files can't be loaded from partitioned directories
**Describe the bug**
A new feature allows creating tables from a set of distributed files resident on each worker filesystem. So, by enabling the flag local_files as True, each worker will try to load their corresponding partitions. However, this feature works fine with CSV files, but not with parquet files where there is a more complex flow due to their row groups.

Create a Dask cluster, where each worker has a disjoint set of parquet files from the same table (make sure the parquet files have more than one row group). Then, run a script in which this new table is created with the flag local_files enabled and tests some queries.

**Expected behavior**
Any output should be generated correctly.

 - Method of cuDF install: From source, nightly branch

rommelDB:
Closes #1062


pyblazing/pyblazing/apiv2/context.py
```diff
@@ -1716,6 +1718,13 @@ def create_table(self, table_name, input, **kwargs):
         input : data source for table.
                 cudf.Dataframe, dask_cudf.DataFrame, pandas.DataFrame,
                 filepath for csv, orc, parquet, etc...
+        file_format (optional) : string describing the file format
+                      (e.g. "csv", "orc", "parquet") this field must
+                      only be set if the files do not have an extension.
+        local_files (optional) : boolean, must be set to True if workers
+                      only have access to a subset of the parquet files
```
williamBlazing, line 1725: `subset of the parquet files` should be `subset of the files`

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -2065,7 +2082,8 @@ def create_table(self, table_name, input, **kwargs):
                 metadata_ids = table.metadata[
                     ["file_handle_index", "row_group_index"]
                 ].to_pandas()
-                grouped = metadata_ids.groupby("file_handle_index")
+
+                grouped = metadata_ids.groupby(metadata_ids["file_handle_index"])
```
williamBlazing, line 2086: This does not make sense. Can you explain this? This used to be groupby a column name, now its group by a column of values?

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -2388,8 +2401,10 @@ def _optimize_skip_data_getSlices(
                 file_and_rowgroup_indices = (
                     file_indices_and_rowgroup_indices.to_pandas()
                 )
-                grouped = file_and_rowgroup_indices.groupby("file_handle_index")
 
+                grouped = file_and_rowgroup_indices.groupby(
```
williamBlazing, line 2405: This does not make sense. Can you explain this? This used to be groupby a column name, now its group by a column of values?

williamBlazing (changes requested):

pyblazing/pyblazing/apiv2/context.py
rommelDB: Fixed.

pyblazing/pyblazing/apiv2/context.py
rommelDB: They were remains of code that I did not remove well, good catch thanks!

PR #1065, opened 2020-10-06
[REVIEW] Remove thrift from conda recipe and build process

kkraus14:
Seems like thrift is no longer used as a dependency in blazingsql so removing it from conda / cmake / docker / etc.


mario21ic (approved):
Looks great to me

Issue #987, opened 2020-09-02
Error messages to connect to S3 and GS


kharoc:
When I execute, for example :

`bc.s3(authority, bucket_name=bucket_name, encryption_type=S3EncryptionType.NONE, access_key_id=access_key_id, secret_key=secret_key)`

The connection is successful but it showed in console the next message. And it seems that there is an error.

` 2020-09-01T11:33:11Z|2110168128|TRACE|deregisterFileSystem: filesystem authority not found. `

felipeblazing:
This may not be an nissue

```
2020-09-01T11:33:11Z|2110168128|TRACE|deregisterFileSystem: filesystem authority not found.

```
Is probably an error message that shows up because it tries to deregister the filesystem in case it existed before this. We could clean it up by not throwing an error if it didnt exist before but its not a big concern

Christian8491:
This was really not an error connection. It was removed fro the PR #1068

Issue #1034, opened 2020-09-23
S3 create table with data from kms bucket from different region.


kharoc:
When I register a kms bucket from a different region of the cluster, I can't create tables with data from this bucket. It happen only with kms bucket even when I send the region as parameter.

PR #1068, opened 2020-10-07
[REVIEW] Just adds a docs important links and avoid the message about filesystem authority not found

Christian8491:
This PR only avoid the logging message about `deregisterFileSystem: filesystem authority not found"` when an authority is sent, Also add a important link related with s3 stuff.


williamBlazing:
Could you please verify that this resolves #987

Christian8491:
Yes. This PR resolves it,

io/src/FileSystem/private/FileSystemManager_p.cpp
```diff
@@ -79,8 +79,6 @@ bool FileSystemManager::Private::deregisterFileSystem(const std::string & author
 	}
 
 	if(found == false) {
```
williamBlazing, line 81: This can just be replaced by `return found;`

williamBlazing (changes requested):

io/src/FileSystem/private/FileSystemManager_p.cpp
Christian8491: updated.

Christian8491:
The issue #1034 should be closed also. As we know the issue was related to key access.

PR #1071, opened 2020-10-08
[REVIEW] Fix for empty folders

### LINKED ISSUE #1070: [BUG] The engine crashes when we try to create a table from an empty directory
**Describe the bug**
The engine crashes when we try to create a table from an empty directory.

The reproducer is as follow:
```
bc = BlazingContext()
bc.create_table('lineitem', '/home/ubuntu/tpch/') # this is an empty dir
```

**Expected behavior**
An error should be raised with no crash.

 - Method of cuDF install: From source branch-0.16

rommelDB:
Fixes #1070

### DIFF
--- a/engine/src/io/data_provider/UriDataProvider.cpp
+++ b/engine/src/io/data_provider/UriDataProvider.cpp
@@ -150,15 +150,28 @@ data_handle uri_data_provider::get_next(bool open_file) {
 				this->directory_uris = BlazingContext::getInstance()->getFileSystemManager()->list(target_uri);
 			}
 
-			std::string ender = ".crc";
+			// sometimes parquet directories have somes files that
+			// have not the same schema as the *.parquet files
+			// we don't want the data provider handle this ones
+			std::vector<std::string> ignored_suffixes{".crc", "_metadata", "_SUCCESS"};
+
 			std::vector<Uri> new_uris;
 			for(int i = 0; i < this->directory_uris.size(); i++) {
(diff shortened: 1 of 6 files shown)


rommelDB:
rerun tests

engine/src/io/data_provider/UriDataProvider.cpp
```diff
@@ -150,18 +150,45 @@ data_handle uri_data_provider::get_next(bool open_file) {
 				this->directory_uris = BlazingContext::getInstance()->getFileSystemManager()->list(target_uri);
 			}
 
-			std::string ender = ".crc";
+			// sometimes parquet directories have somes files that
+			// have not the same schema as the *.parquet files
+			// we don't want the data provider handle this ones
+			std::vector<std::string> ignored_suffixes{".crc", "_metadata", "_SUCCESS"};
```
felipeblazing, line 156: we should add ipython notebook checkpoints to this those get made everywhere

felipeblazing (approved):

engine/src/io/data_provider/UriDataProvider.cpp
felipeblazing: .ipynb_checkpoints
are the files itmakmes

romeluko: Added. Thank you.

rommelDB: Added, thank you.

Issue #1087, opened 2020-10-19
Unit tests for UriDataProvider


williamBlazing:
We want to build unit tests for the UriDataProvider. 

The unit tests will need to create folders and test files in a tmp location for the unit tests to run.

The test suite should contain for following tests:
Test for empty folder
Test for folder with one file
Test for folder with multiple files
Test for folder with multiple files and one empty folder 
Test for folder with multiple files and one non-empty folder 
Test for folder with multiple folder, each with a couple files 
Test for wildcard that returns nothing 
Test for wildcard that returns one file 
Test for wildcard that returns multiple files 

For each of these we should test with and without ignore_missing_paths = True

Another set of tests should include setting `std::vector<std::map<std::string, std::string>> uri_values`

williamBlazing:
For the set of tests that use `std::vector<std::map<std::string, std::string>> uri_values`
We want to always run it with  ignore_missing_paths = True
We will want to test the following:
one folder with multiple files, one folder with one file and one folder with no files

diegodfrf:
Implemented in PR #1090

PR #1090, opened 2020-10-21
[REVIEW] Adding additional provider test

diegodfrf:
Bug #1087

### DIFF
--- a/engine/tests/provider/provider_test.cpp
+++ b/engine/tests/provider/provider_test.cpp
@@ -533,6 +533,8 @@ TEST_F(ProviderTest, wildcard_multiple_file)
 	}
 
 	std::vector<std::string> cmp{BLAZING_TMP_PATH + "/filetwo.orc", BLAZING_TMP_PATH + "/filethree.orc"};
+	std::sort(cmp.begin(), cmp.end());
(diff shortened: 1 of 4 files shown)


GPUtester:

engine/tests/provider/provider_test.cpp
```diff
@@ -37,16 +37,39 @@ void create_dummy_file(std::string content, std::string filename){
 	outfile.close();
 }
 
+void create_folder_test()
```
felipeblazing, line 40: Just because the api returns true or false doe snot mean the directory was properly created. You need to make sure it actually exists using file apis

felipeblazing (changes requested):

williamBlazing:
Add to whitelist

engine/tests/provider/provider_test.cpp
diegodfrf: Done

williamBlazing:
rerun-tests

rommelDB:
rerun-tests

rommelDB:
rerun tests

felipeblazing (approved):

PR #1091, opened 2020-10-21
[REVIEW] Improves the error message when validating the GCP bucket

rommelDB:
This PR improves the error returned by the Google API showing more information about the origin of the problem when validating any bucket.


williamBlazing:
rerun-tests

PR #1099, opened 2020-10-26
[REVIEW] fixed issue with config_options

williamBlazing:


williamBlazing:
rerun-tests

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -1309,11 +1309,11 @@ def __init__(
         # want to use config_options and not self.config_options
         # since its not encoded
         if "BLAZING_LOGGING_DIRECTORY" in self.config_options:
```
rommelDB, line 1311: Due to load_config_options_from_env function returns the encoded config_options,
the variable name should be encoded too, otherwise it will have no effect:

For example:
`if "BLAZING_LOGGING_DIRECTORY".encode() in self.config_options:`

williamBlazing: good call.

Christian8491:
Please, before merge in this PR try in a distributed cluster, or let me know to try this.

PR #1105, opened 2020-10-30
[REVIEW] Implement to_date/to_timestamp functions

### LINKED ISSUE #428: Implement TO_DATE(string_col, format_string)
We currently can CAST from a string to a date, but that assumes only one format, we want to be able to do the same, but specifying the format like this: http://www.sqlines.com/oracle-to-sql-server/to_date . This is not part of the SQL standard that Calcite uses out of the box, but it is fairly standard and its my understanding that Calcite does support this, but its not enabled by default. We need to figure out how to enable this syntax in Calcite. 

Then in the engine we can do something similar to what we do here: https://github.com/BlazingDB/blazingsql/blob/branch-0.17/engine/src/execution_graph/logic_controllers/LogicalProject.cpp#L362  but providing more flexibility by allowing to specify the `format_string` in
`cudf::strings::to_timestamps(column, cudf::data_type{cudf::type_id::TIMESTAMP_NANOSECONDS}, format_string);`

This feature would require adding it to our readme.io docs and also require new end 2 end tests

jeanp413:
Fixes #428


jeanp413:
rerun tests

felipeblazing (approved):

PR #1108, opened 2020-11-02
[REVIEW] Always remove temp files

### LINKED ISSUE #1107: Check to make sure that Clean up blazing_cache on execution startup works

jeanp413:
Fixes #1107


pyblazing/pyblazing/apiv2/context.py
```diff
@@ -829,15 +829,12 @@ def get_current_directory_path():
...
+    if os.path.isdir(data_dir):  # only if data_dir exists
         all_files = os.listdir(data_dir)
-        current_time = time.time()
         for file in all_files:
             if ".blazing-temp" in file:
                 full_path_file = data_dir + "/" + file
-                creation_time = os.path.getctime(full_path_file)
-                if (current_time - creation_time) // (1 * 60 * 60) >= 1:
-                    os.remove(full_path_file)
+                os.remove(full_path_file)
```
williamBlazing, line 837: we want to leave files that are not that old because if you are running a query and another context gets created, it could delete files from another query that is running

williamBlazing (changes requested):

pyblazing/pyblazing/apiv2/context.py
felipeblazing: We cant have multiple contexts at the same time anyway how is this a problem?

williamBlazing (changes requested):

williamBlazing:
rerun tests

Issue #1106, opened 2020-10-31
[BUG] Jupyter - Failed opening file blazing_log/RAL.0.log for writing: Permission denied


lmeyerov:
**Describe the bug**

In a notebook environment, we get error `Failed opening file blazing_log/RAL.0.log for writing: Permission denied`, though `cudf` imports + computes fine

**Steps/Code to reproduce bug**

```
from blazingsql import BlazingContext
bc = BlazingContext()
```

Yields 

```
---------------------------------------------------------------------------
InitializeError                           Traceback (most recent call last)
<ipython-input-18-7813b5513ccc> in <module>
      1 from blazingsql import BlazingContext
----> 2 bc = BlazingContext()

/conda/envs/rapids/lib/python3.7/site-packages/pyblazing/apiv2/context.py in __init__(self, dask_client, network_interface, allocator, pool, initial_pool_size, maximum_pool_size, enable_logging, config_options)
   1423                 enable_logging=enable_logging,
   1424                 config_options=self.config_options,
-> 1425                 logging_dir_path=logging_dir_path,
   1426             )
   1427             node = {}

/conda/envs/rapids/lib/python3.7/site-packages/pyblazing/apiv2/context.py in initializeBlazing(ralId, networkInterface, singleNode, allocator, pool, initial_pool_size, maximum_pool_size, enable_logging, config_options, logging_dir_path)
    208         initial_pool_size,
    209         maximum_pool_size,
--> 210         enable_logging,
    211     )
    212 

/conda/envs/rapids/lib/python3.7/site-packages/bsql_engine/io/io.pyx in cio.initializeCaller()

/conda/envs/rapids/lib/python3.7/site-packages/bsql_engine/io/io.pyx in cio.initializeCaller()

/conda/envs/rapids/lib/python3.7/site-packages/bsql_engine/io/io.pyx in cio.initializePython()

InitializeError: [Initialize Error] Failed opening file blazing_log/RAL.0.log for writing: Permission denied

```

Also of interest, from Python 3 in the container running jupyter,  (variant of https://hub.docker.com/r/graphistry/graphistry-forge-base with `source activate rapids`), we see:

```
(rapids) graphistry@f7f4c4bf923e:~$ python
Python 3.7.7 (default, Mar 26 2020, 15:48:22) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
>>> 
[...]

PR #1110, opened 2020-11-02
[REVIEW] Feature adding local logging dirrectory

### LINKED ISSUE #1083: [FEA] Optionally setting a Blazing local logging directory
**Is your feature request related to a problem? Please describe.**
If we run a script with a remote dask client attached to a Blazing Context and we pass a path to the  BLAZING_LOGGING_DIRECTORY config option. An error will raised if the folder passed don't exists on the client machine. A workaround consists of creating the same folder locally. But, it will depend on whether we have the required privileges to make it.

**Describe the solution you'd like**
If we are not able to create a folder, let's say because we lack privileges. We should pass another config option, for example, BLAZING_LOCAL_LOGGING_DIRECTORY pointing to a folder locally accessible.

rommelDB:
Fixes #1083 and #1106


tests/BlazingSQLTest/EndToEndTests/loggingTest.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def executionTest():
+        # Run Query ------------------------------------------------------
+
+        print("==============================")
+        print(queryType)
+        print("==============================")
+
+        queryId = "TEST_01"
+        query = """select * from bsql_logs"""
```
williamBlazing, line 22: Please change this query by adding a filter. After running thousands of queries, the return of this select * could be enormous.

williamBlazing (changes requested):

tests/BlazingSQLTest/EndToEndTests/loggingTest.py
rommelDB: Fixed. Thanks.

williamBlazing:
This also fixes #1106

williamBlazing:
rerun-tests

PR #1111, opened 2020-11-02
adding output of json files for cppcheck

felipeblazing:
added           -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \ to the cmake of io, comms, and engine cpp libraries so we can analyze with cppcheck


felipeblazing:
rerun tests

felipeblazing:
rerun tests

PR #1143, opened 2020-11-10
Updating branch-0.17 with everything merged into main (previously master)

williamBlazing:


williamBlazing:
this PR should not be done from main. this is the wrong way to do this

Issue #1123, opened 2020-11-06
[BUG] BlazingContext.partition() throws a RuntimeWarning


drabastomek:
**Describe the bug**
When run in a `python` script (not in a notebook), the `.repartition(...)` method throws a `RuntimeWarning` that `Client.close()` was never awaited.

**Steps/Code to reproduce bug**
Here's a minimal example to run from CLI: `python test.py`. 

`test.py`:

```
import cudf
import dask_cudf
import blazingsql as blazing
from dask_cuda import LocalCUDACluster
from dask.distributed import Client
​
if __name__ == '__main__':
    df = dask_cudf.from_cudf(cudf.DataFrame([
        (2,3,4,5)
        , (1,4,2,5)
        , (9,2,45,1)
        , (4,2,4,1)
        , (4,1,4,1)
    ], columns=['a','b','c','d']), npartitions=2)
​
    cluster = LocalCUDACluster()
    client = Client(cluster)
​
    bc = blazing.BlazingContext(dask_client=client)
    bc.create_table('test', df)
    fff = bc.sql('SELECT * FROM test')
    fff = bc.partition(fff, by=['a', 'b'])
​
    fff['a'] = fff.map_partitions(lambda df: df['a'].astype('float64'))
```

**Expected behavior**
The script finished without any warnings.

**Environment overview (please complete the following information)**
 - Environment location: Docker
 - Method of cuDF install: Docker
   - Docker pull from stable, v.16, CUDA 11

williamBlazing:
This issue is related #1076  https://github.com/BlazingDB/blazingsql/issues/1076

williamBlazing:
Closing this issue since it will be addressed in other issues

PR #1158, opened 2020-11-16
deprecating bc.partition

williamBlazing:
We can use `ddf.shuffle(on=[colnames])` instead of bc.partition, so we should just deprecate it.

This closes #1011  and #1123


williamBlazing:
rerun-tests

PR #1187, opened 2020-11-19
[REVIEW] Enable MySQL-specific SQL operators in addition to Standard and Oracle

### LINKED ISSUE #1178: [FEA] Enable dialect-specific operators with CalciteConnectionProperty
Calcite [supports](https://calcite.apache.org/docs/reference.html#dialect-specific-operators) a variety of useful operators that are not in the SQL standard.

It appears these can be enabled by setting the `FUN` parameter of CalciteConnectionProperty for the connection.

https://calcite.apache.org/docs/adapter.html#jdbc-connect-string-parameters

This is necessary for:
- https://github.com/BlazingDB/blazingsql/issues/1173
- https://github.com/BlazingDB/blazingsql/issues/1174
- https://github.com/BlazingDB/blazingsql/issues/1176

beckernick:
This PR:
- Enables MySQL-specific SQL operators, which makes several useful string operators potentially available (REVERSE, LEFT, RIGHT, etc.). Currently only Standard SQL and Oracle SQL are available.

I've confirmed that this change does allow Calcite to use these operators. Passes `./test.sh` locally.

This closes #1178 

```python
import pandas as pd

bc = BlazingContext()​
​
df = pd.DataFrame({
    "a": ["one", "two"],
    "b": [1, 2],
})
​
bc.create_table("df", df)
​
query = """
SELECT
    a,
    REVERSE(a)
FROM df
"""

print(bc.explain(query))
LogicalProject(a=[$0], EXPR$1=[REVERSE($0)])
  BindableTableScan(table=[[main, df]], projects=[[0]], aliases=[[a, EXPR$1]])

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - #1158 Deprecated bc.partition
 - #1154 Recompute the avg_bytes_per_row value
 - #1155 Removing comms subproject and cleaning some related code
+- #1187 Enable MySQL-specific SQL operators in addition to Standard and Oracle
 
 
 ## Bug Fixes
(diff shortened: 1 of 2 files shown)


felipeblazing:
@beckernick have you checked to see that you can actually run those queries. i do not believe we have reverse implemented as a function so that would need to be done first.

beckernick:
> @beckernick have you checked to see that you can actually run those queries. i do not believe we have reverse implemented as a function so that would need to be done first.

Yep, we're aligned. That will be implemented in a separate PR that closes #1173

williamBlazing:
Ideally we should see if there are any operators that are implemented in the different SQL standards that have the same keyword, but with different APIs. I am not sure if we know of any in particular but if we do, it would be useful to know how Calcite would handle this situation

williamBlazing:
rerun-tests

PR #1188, opened 2020-11-19
[REVIEW] Implement UPPER and LOWER operators and tests

### LINKED ISSUE #1134: [FEA] Support string UPPER
I'd like to be able to call UPPER on a string column to convert the case to uppercase. The alternative today would be to go to cudf/dask and use `str.upper()`, which can be difficult inside query logic.

```python
from pyspark.sql import SparkSession
​
spark = SparkSession.builder \
    .master("local") \
    .getOrCreate()
​
spark.sql("SELECT UPPER('Example')").show()
+--------------+
|upper(Example)|
+--------------+
|       EXAMPLE|
+--------------+
```

### LINKED ISSUE #1135: [FEA] Support string LOWER
I'd like to be able to call LOWER on a string column to convert the case to lowercase. The alternative today would be to go to cudf/dask and use `str.lower()`, which can be difficult inside query logic.

```python
from pyspark.sql import SparkSession
​
spark = SparkSession.builder \
    .master("local") \
    .getOrCreate()
​
spark.sql("SELECT LOWER('Example')").show()
+--------------+
|lower(Example)|
+--------------+
|       example|
+--------------+
```

ChrisJar:
Enables use of UPPER or LOWER in a query and adds end to end tests for both operators

Passes ./test.sh locally.

This closes #1134 and closes #1135

### DIFF
--- a/engine/src/parser/expression_utils.cpp
+++ b/engine/src/parser/expression_utils.cpp
@@ -52,6 +52,8 @@ bool is_unary_operator(operator_type op) {
 	case operator_type::BLZ_CAST_TIMESTAMP:
 	case operator_type::BLZ_CAST_VARCHAR:
 	case operator_type::BLZ_CHAR_LENGTH:
+	case operator_type::BLZ_STR_LOWER:
+	case operator_type::BLZ_STR_UPPER:
 		return true;
 	default:
 		return false;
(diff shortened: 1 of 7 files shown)


felipeblazing (approved):
This is great! Waiting on gpuci but if things look good we can merge

felipeblazing:
rerun tests

felipeblazing:
rerun tests

ChrisJar:
rerun tests

engine/src/parser/expression_utils.cpp
```diff
@@ -220,6 +222,10 @@ cudf::type_id get_output_type(operator_type op, cudf::type_id input_left_type, c
 	case operator_type::BLZ_STR_SUBSTRING:
 	case operator_type::BLZ_STR_CONCAT:
 		return cudf::type_id::STRING;
+	case operator_type::BLZ_STR_LOWER:
+		return cudf::type_id::STRING;
+	case operator_type::BLZ_STR_UPPER:
+		return cudf::type_id::STRING;
```
beckernick, line 228: would it make sense to add these cases under the `return cudf::type_id::STRING;` above?

ChrisJar: Yep, good point. Just fixed

felipeblazing:
rerun tests

felipeblazing:
rerun tests

PR #1196, opened 2020-11-20
[WIP] Fix colum_names always as list of str

### LINKED ISSUE #1190: [BUG] bc.describe_table crashes for cudf and dask_cudf tables
Running `bc.describe_table()` for a table that is from a `cudf` or `dask_cudf` dataframe fails.

This is because the variable `column_names`  for the table is a string, where as for a file table `column_names` is in bytes (encoded string).

We should fix this by making sure that `column_names` is always a string. We need to make sure to fix this in the `describe_table` function and anywhere we use that variable.

Christian8491:
This PR closes #1190


pyblazing/pyblazing/apiv2/context.py
```diff
@@ -2193,6 +2193,9 @@ def create_table(self, table_name, input, **kwargs):
                     row_groups_ids.append(row_group_ids)
                 table.row_groups_ids = row_groups_ids
 
+            # want all `column_names` be a list of strings
+            table.column_names = [i.decode() for i in table.column_names]
```
williamBlazing, line 2197: This should be handled at the source. Where is column_names being set originally? Please fix this at the source.

williamBlazing (changes requested):

PR #1200, opened 2020-11-23
[REVIEW] Implement string REGEXP_REPLACE

### LINKED ISSUE #1195: [FEA] Support REGEXP_REPLACE for strings
I'd like to be able to use REGEXP_REPLACE to replace specific characters based on a regular expression in every string of a column. This is listed as a supported operation on string columns in the Calcite reference, but is currently not implemented.

```python
from pyspark.sql import SparkSession
import pandas as pd
​
​
# spark = SparkSession.builder \
#     .master("local") \
#     .getOrCreate()
​
# bc = BlazingContext()
​
​
df = pd.DataFrame({
    "a": ["Felipe", "William", "Rodrigo"],
    "b": [2, 4, 6],
    "c": ["2020-11-20", "2020-11-19", "2020-11-18"]
})
​
bc.create_table("df", df)
sdf = spark.createDataFrame(df)
sdf.createOrReplaceTempView("df")
​
query = """
SELECT
    a,
    REGEXP_REPLACE(a, '[aio]', '#')
FROM df
"""
​
spark.sql(query).show()
​
print(bc.explain(query))
# print(bc.sql(query))
+-------+---------------------------+
|      a|regexp_replace(a, [aio], #)|
+-------+---------------------------+
| Felipe|                     Fel#pe|
|William|                    W#ll##m|
|Rodrigo|                    R#dr#g#|
+-------+---------------------------+

LogicalProject(a=[$0], EXPR$1=[REGEXP_REPLACE($0, '[aio]', '#')])
  BindableTableScan(table=[[main, df]], projects=[[0]], aliases=[[a, EXPR$1]])
```

beckernick:
This PR:
- [x] Implements the REGEXP_REPLACE operator for string columns using cudf::strings::replace_with_backrefs (and cudf::strings::slice_strings and cudf::strings::concatenate to handle the optional `pos` argument)
- [x] Adds a new end-to-end test in stringsTests.py
- [x] Passes tests locally
- [x] PR to update testing files

Note that the `position` argument is 1-indexed (rather than 0 indexed) to be consistent with Oracle and MySQL.

This closes #1195 

```python
from pyspark.sql import SparkSession
import pandas as pd


# spark = SparkSession.builder \
#     .master("local") \
#     .getOrCreate()

# bc = BlazingContext()


df = pd.DataFrame({
    "a": ["555.456.7890", "555.456.1234", "555.123.7890"],
    "b": [2, 4, 6],
})

bc.create_table("df", df)

query = """
SELECT
    a,
    REGEXP_REPLACE(a, '([0-9]{3})\.([0-9]{3})\.([0-9]{4})', '\\2 \\3-\\1') as clean
FROM df
"""
​
print(bc.sql(query))
              a         clean
0  555.456.7890  456 7890-555
1  555.456.1234  456 1234-555
2  555.123.7890  123 7890-555
```

MySQL Example:
```
SET @str = 'Cat Dog Cat Dog Cat';
SELECT 
  @str 'Original String',
  REGEXP_REPLACE(@str, 'Cat', 'Tiger', 2) 'Result';
Result:

+---------------------+-------------------------+
| Original String     | Result                  |
+---------------------+-------------------------+
| Cat Dog Cat Dog Cat | Cat Dog Tiger Dog Tiger |
+---------------------+-------------------------+
```

```python
bc.create_table("df_slice", df.iloc[2:])
​
query = """
SELECT
    a,
    REGEXP_REPLACE(a, 'Cat', 'Tiger', 2) as clean
FROM df_slice
"""
​
print(bc.sql(query))
                     a                    clean
0  Cat Dog Cat Dog Cat  Cat Dog Tiger Dog Tiger
```


GPUtester:

engine/src/execution_graph/logic_controllers/LogicalProject.cpp
```diff
@@ -132,6 +133,39 @@ std::unique_ptr<cudf::column> evaluate_string_functions(const cudf::table_view &
...
+
+            std::unique_ptr<cudf::column> prefix_col = cudf::strings::slice_strings(column, prefix, start);
+            cudf::column_view post_replace_view = cudf::strings::replace_with_backrefs(
+                cudf::column_view(cudf::strings::slice_strings(column, start)->view()),
+                pattern,
+                repl)->view();
+
+            computed_col = cudf::strings::concatenate(
+                cudf::table_view{{prefix_col->view(), post_replace_view}}
+            );
```
beckernick, line 163: The logic for this section is:
- If we have a `position` argument, we should only replace within the string starting from index `position`

To make this concrete, see the following MySQL example (note the indexing starts at 1). By default we replace all matches. With the position argument, we only replace valid matches starting from position 3 in this case, but still return the parts of the string before the we look for matches.

```
mysql> SELECT REGEXP_REPLACE('abc def ghi', '[a-z]+', 'X');
REGEXP_REPLACE('abc def ghi', '[a-z]+', 'X')
X X X
```
```mysql>SELECT REGEXP_REPLACE('abc def ghi', '[a-z]+', 'X', 3);
REGEXP_REPLACE('abc def ghi', '[a-z]+', 'X', 3)
abX X X
```

@davidwendt , do you have any thoughts on if there is a better way to do this than I have here?
- Slice the strings to get the characters before the `position` argument
- Slice the strings starting at the `position` argument and pass that to replace_with_backrefs
- Concatenate the results, recombining the "prefix" (portions of strings not sued for replace) with the "replaced" segment

beckernick:
Retargeting to 0.18

engine/src/execution_graph/logic_controllers/LogicalProject.cpp
davidwendt: The logic looks right. I don't see a better way to do it. Maybe if we had a _split_on_position function_ you would not have to slice the strings twice (left and right).

The code has an error though. The `concatenate()` call may crash. You need to keep the `post_replace_view's` column alive until after `concatenate()` returns. Changing it to following would be safer:
```
        auto prefix_col = cudf::strings::slice_strings(column, prefix, start);
        auto post_replace_col = cudf::strings::replace_with_backrefs(
            cudf::column_view(cudf::strings::slice_strings(column, start)->view()),
            pattern,
            repl);

        computed_col = cudf::strings::concatenate(
            cudf::table_view{{prefix_col->view(), post_replace_col->view()}} );
```

beckernick: Thanks for confirming the logic, and for identifying and proposing a solution to the risky segment @davidwendt!

beckernick: Resolved by https://github.com/BlazingDB/blazingsql/pull/1200/commits/06ad76d9fbb11fd3f7b2746183c49c28b4cae612

beckernick:
There are two aspects of this functionality that cannot be cleanly tested using Spark or Drill with the current pattern of e2e tests:
- Backreferences in the regex replacement string use `$` in Spark/Drill but `\\` in Calcite and Oracle
- The `position` argument available in Calcite, Oracle, and MySQL is not available in Spark or Drill

In both scenarios, we can't send the same query to Blazing as Spark/Drill. Perhaps it may be worth defining a process for how to handle this, as it may come up in the future as well. One suggestion from @williamBlazing was that we might send different queries to different engines based on the expected output. With this, it would be possible to address both aspects.

It looks like that would also require updates to the testing logic, as right now the engines get sent the same query by design. If we go down that route, would people prefer :
- This PR only includes tests for the basic functionality, with a follow-up PR to allow for more varied testing that will also add tests for backreferences and the position argument
- Do it all in this PR

I would vote for only including basic functionality in this PR, but defer to others (cc @williamBlazing @Christian8491 )

rommelDB:
If I'm not mistaken, In the E2E tests there is currently a way to execute slightly different queries that produce the same result depending on the engine. This is an example:
https://github.com/BlazingDB/blazingsql/blob/84f61092d9b625eba21efae4c87b25f5b95c9507/tests/BlazingSQLTest/EndToEndTests/timestampTest.py#L158

cc @beckernick @williamBlazing

beckernick:
Ah, fantastic. Thanks for the pointer  @rommelDB . I didn't catch it while scanning through 👍

beckernick:
PR for testing data: https://github.com/BlazingDB/blazingsql-testing-files/pull/9

Marking as ready for review and updating CHANGELOG

beckernick:
CI failing due to https://github.com/BlazingDB/blazingsql-testing-files/pull/9 putting the files in the wrong location. https://github.com/BlazingDB/blazingsql-testing-files/pull/10 will resolve it.

beckernick:
rerun tests

beckernick:
Style checks are failing on files not touched in this PR:

```
pyblazing/pyblazing/apiv2/context.py:2581:1: W293 blank line contains whitespace
pyblazing/pyblazing/apiv2/context.py:2586:1: W293 blank line contains whitespace
pyblazing/pyblazing/apiv2/context.py:2588:9: E303 too many blank lines (2)
pyblazing/pyblazing/apiv2/context.py:2673:1: W293 blank line contains whitespace
pyblazing/pyblazing/apiv2/context.py:2675:5: E303 too many blank lines (2)
```

beckernick:
rerun tests

PR #1207, opened 2020-11-25
[REVIEW] Calcite uses literal as int32 if  not explicit CAST was provided

### LINKED ISSUE #1041: Issue with cast literal as INT
doing something as `cast(32 as INT)` outputs a column of type int8 when it should be int32

Christian8491:
This PR closes #1041 

By default Calcite uses `int32` for literal values (as long as it does not exceed the limit of an `int32`) 
For each of these queries 
```
query = "select cast(32 as INT) as col_a from df"
query = "select cast(32 as INTEGER) as col_a from df"  
query = "select 32 as col_a from df"  
```

The Relational Algebra returns:
```
DEBUG: com.blazingdb.calcite.application.RelationalAlgebraGenerator - non optimized
LogicalProject(col_a=[32])
  LogicalTableScan(table=[[main, df]])

DEBUG: com.blazingdb.calcite.application.RelationalAlgebraGenerator - optimized
LogicalProject(col_a=[32])
  BindableTableScan(table=[[main, df]], projects=[[]], aliases=[[col_a]])
```

Some of other SQL engines (as `pyspark.sql`) returns also a `int32`, at least you explicitly apply  a `CAST`

### DIFF
--- a/engine/src/parser/expression_tree.cpp
+++ b/engine/src/parser/expression_tree.cpp
@@ -227,16 +227,8 @@ cudf::data_type infer_type_from_literal_token(const lexer::token & token) {
       return parsed_double == casted_float ? cudf::data_type{cudf::type_id::FLOAT32} : cudf::data_type{cudf::type_id::FLOAT64};
     } else {
       int64_t parsed_int64 = std::stoll(token_value);
-      return parsed_int64 > std::numeric_limits<int32_t>::max()
-              || parsed_int64 < std::numeric_limits<int32_t>::min()
-                ? cudf::data_type{cudf::type_id::INT64}
-                : parsed_int64 > std::numeric_limits<int16_t>::max()
-                  || parsed_int64 < std::numeric_limits<int16_t>::min()
-                      ? cudf::data_type{cudf::type_id::INT32}
-                      : parsed_int64 > std::numeric_limits<int8_t>::max()
-                        || parsed_int64 < std::numeric_limits<int8_t>::min()
-                          ? cudf::data_type{cudf::type_id::INT16}
-                          : cudf::data_type{cudf::type_id::INT8};
+      // as other SQL engines, defaults to int32
+      return cudf::data_type{cudf::type_id::INT32};
     }
   } else if(token.type == lexer::token_type::Timestamp) {
     const std::string & token_value = token.value;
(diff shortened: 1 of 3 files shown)


engine/src/parser/expression_tree.cpp
```diff
@@ -227,16 +227,8 @@ cudf::data_type infer_type_from_literal_token(const lexer::token & token) {
...
-                ? cudf::data_type{cudf::type_id::INT64}
-                : parsed_int64 > std::numeric_limits<int16_t>::max()
-                  || parsed_int64 < std::numeric_limits<int16_t>::min()
-                      ? cudf::data_type{cudf::type_id::INT32}
-                      : parsed_int64 > std::numeric_limits<int8_t>::max()
-                        || parsed_int64 < std::numeric_limits<int8_t>::min()
-                          ? cudf::data_type{cudf::type_id::INT16}
-                          : cudf::data_type{cudf::type_id::INT8};
+      // as other SQL engines, defaults to int32
+      return cudf::data_type{cudf::type_id::INT32};
```
williamBlazing, line 231: I think it should be:
```
return parsed_int64 > std::numeric_limits<int32_t>::max() 
              || parsed_int64 < std::numeric_limits<int32_t>::min()
                   ? cudf::data_type{cudf::type_id::INT64} : cudf::data_type{cudf::type_id::INT32};
```
so that if the number is bigger than what can fit in an int32 its treated accordingly, and only otherwise do we just treat it as a default of int32

williamBlazing (changes requested):

engine/src/parser/expression_tree.cpp
Christian8491: Fortunately for us, Calcite is smart enough and if the literal value es larger than a `int32`, then it applies `CAST` to a `BIGINT` in its Relational Algebra, so for this query:

`query = "select 3255455842 as col_a from df"  `

The Relational Algebra is:
```
DEBUG: com.blazingdb.calcite.application.RelationalAlgebraGenerator - non optimized
LogicalProject(col_a=[3255455842:BIGINT])
  LogicalTableScan(table=[[main, df]])

DEBUG: com.blazingdb.calcite.application.RelationalAlgebraGenerator - optimized
LogicalProject(col_a=[3255455842:BIGINT])
  BindableTableScan(table=[[main, df]], projects=[[]], aliases=[[col_a]])
```

williamBlazing:
rerun tests

PR #1224, opened 2020-12-03
added cudaSetDevice to thread initialization so the context is availa…

felipeblazing:
…ble when ucx tries to ipc things


GPUtester:

PR #1232, opened 2020-12-04
[REVIEW] Update join kernel to use the task executor

### LINKED ISSUE #1215: Updating PartwiseJoin kernel to use the new task executor
This issue is to discus how we should update the PartwiseJoin kernel to use the task executor.
Currently there is blocker as the PartwiseJoin kernel uses `leftArrayCache` and `rightArrayCache` to store the left and right batches that will be joined but the task executor also keeps needs ownership of the input batches when we call `add_task`.
The simplest option would be to clone the batches.
```c++
std::vector<std::unique_ptr<ral::cache::CacheData>> inputs;
inputs.push_back(std::move(left_cache_data));
inputs.push_back(std::move(right_cache_data));

ral::execution::executor::get_instance()->add_task(
						std::move(inputs),
						this->output_cache(),
						this);

mark_set_completed(left_ind, right_ind);
```

jeanp413:
Fixes #1215


GPUtester:

jeanp413:
rerun tests

Issue #1221, opened 2020-12-02
[BUG] RAL_FAIL exceptions should be caught upstream to avoid segfaults


beckernick:
It looks like hitting a RAL_FAIL throws the error and causes a segfault. It would be nice to catch situations that lead to a RAL_FAIL in the Python layer (or wherever appropriate) to avoid this segfault.

```python
~/NVIDIA/blazingsql$ ipython
Python 3.7.9 (default, Aug 31 2020, 12:42:55) 
Type "copyright", "credits" or "license" for more information.

IPython 5.8.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import pandas as pd
   ...: 
   ...: bc = BlazingContext()
   ...: 
   ...: 
   ...: df = pd.DataFrame({
   ...:     "a": ["Felipe", "William", "Rodrigo"],
   ...:     "b": [10, 9, 8],
   ...: })
   ...: 
   ...: bc.create_table("df", df)
   ...: 
   ...: query = """
   ...: SELECT SUBSTRING('hello', 0, 2)
   ...: FROM df
   ...: """
   ...: 
   ...: print(bc.sql(query))
   ...: 
BlazingContext ready
[12:30:42.504] [error] 293879500|0|0|In Projection kernel batch 0 for LogicalProject(EXPR$0=['h':VARCHAR(5)]). What: Ral failure at: /home/nicholasb/NVIDIA/blazingsql/engine/src/parser/expression_tree.cpp:283: Invalid literal cast type|||||
[12:30:42.504] [error] 293879500|||In execute_graph. What: Ral failure at: /home/nicholasb/NVIDIA/blazingsql/engine/src/parser/expression_tree.cpp:283: Invalid literal cast type|||||
[nicholasb-HP-Z8-G4-Workstation:120627:0:120627] Caught signal 11 (Segmentation fault: address not mapped to object at address 0x20)
==== backtrace (tid: 120627) ====
 0  /home/nicholasb/miniconda3/envs/bsql/lib/./libucs.so.0(ucs_handle_error+0x12d) [0x7f0f9074850d]
 1  /home/nicholasb/miniconda3/envs/bsql/lib/./libucs.so.0(+0x23951) [0x7f0f90748951]
[...]

PR #1255, opened 2020-12-11
[REVIEW] Added retry mechanism to all the currently implemented tasks

### LINKED ISSUE #1245: Fix how the retry mechanism works in the executor
Right now this is what we have:
```
void kernel::process(std::vector<std::unique_ptr<ral::cache::CacheData > > & inputs,
		std::shared_ptr<ral::cache::CacheMachine> output,
		cudaStream_t stream,
        std::string kernel_process_name = ""){
    std::vector< std::unique_ptr<ral::frame::BlazingTable> > input_gpu;

    if (this->has_limit_ && output->get_num_rows_added() >= this->limit_rows_) {
  //      return;
    }

    for(auto & input : inputs){
        try{
            //if its in gpu this wont fail
            //if its cpu and it fails the buffers arent deleted
            //if its disk and fails the file isnt deleted
            //so this should be safe
            input_gpu.push_back(std::move(input->decache()));

        }catch(std::exception e){
            throw e;
        }
    }

    try{
       do_process(std::move(input_gpu),output,stream, kernel_process_name);
    }catch(std::exception e){
        //remake inputs here
        int i = 0;
        for(auto & input : inputs){
            if (input->get_type() == ral::cache::CacheDataType::GPU || input->get_type() == ral::cache::CacheDataType::GPU_METADATA){
                //this was a gpu cachedata so now its not valid
                static_cast<ral::cache::GPUCacheData *>(input.get())->set_data(std::move(input_gpu[i]));
            }
            i++;
        }
        throw;
    }

}
```

We cant get the data from `input_gpu[i]` in the catch statement so what we have does not work.

What we want to do instead is have `do_process` return a struct `task_output`.
task_output will have three members:
status = enum with options: 'Success', 'Recoverable_fail', 'Fatal_fail'
error = std::string with the error message
inputs = std::vector< std::unique_ptr<ral::frame::BlazingTable> > inputs

Inside each `do_process` function, we will have a try catch encapsulating the whole function. If the function succeeds then
status = Success
error = ""
inputs = empty vector
If there is an exception, then:
status = depending on the exception we will set Recoverable_fail or Fatal_fail
error = the error message of the exception
inputs = if Recoverable_fail then it will contain the original inputs, so that we can get them back for a retry, otherwise empty vector.

With this paradigm change  then we can update `kernel::process` accordingly so that after `do_process ` we check the status and respond accordingly:
- status == Sucess => carry on
- status == Recoverable_fail => re-set the data into the input CacheData
- status == Throw the error up the stack

Additionally I think we should combine what kernel::process does into task::run. This way, we can more easily handle the distinction between Recovarable_fail and Fatal_fail. This will make more sense once the above is implemented.

felipeblazing:
Every do_process function now needs to return a `ral::execution::task_result` so that we can properly recover resources. 

This actually shows one of the biggest weaknesses of using things like exception for error handling. Namely that when the exceptions have resources that can be recovered through different ways and which sometimes have to be moved around to avoid copies we end up going back to something like C style error codes.

Fixes #1245 #1264 #1221


GPUtester:

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.cpp
```diff
@@ -110,67 +116,83 @@ DistributeAggregateKernel::DistributeAggregateKernel(std::size_t kernel_id, cons
...
-                true, //specific_cache
-                "", //cache_id
-                {this->context->getMasterNode().id()}); //target_id
+                send_message(std::move(input),
+                    true, //specific_cache
+                    "", //cache_id
+                    {this->context->getMasterNode().id()}); //target_id
+            }
+        }catch(rmm::bad_alloc e){
+            return {ral::execution::task_status::RETRY, std::string(e.what()), std::vector< std::unique_ptr<ral::frame::BlazingTable> > ()};
```
felipeblazing, line 154: how are you able to retry if you cannot return the inputs. Can we not make send_message just return this after its done?

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.cpp
```diff
@@ -525,39 +556,45 @@ LimitKernel::LimitKernel(std::size_t kernel_id, const std::string & queryString,
...
+                            "output_num_rows"_a=log_output_num_rows,
+                            "output_num_bytes"_a=log_output_num_bytes,
+                            "event_type"_a="compute",
+                            "timestamp_begin"_a=eventTimer.start_time(),
+                            "timestamp_end"_a=eventTimer.end_time());
+
+            output->addToCache(std::move(input));
+        }
+    }catch(rmm::bad_alloc e){
+        return {ral::execution::task_status::RETRY, std::string(e.what()), std::vector< std::unique_ptr<ral::frame::BlazingTable> > ()};
```
felipeblazing, line 593: this is a particularly interesting kernel. It can modify the inputs BUT applying limit_table more than once should NOT have negative consequences. Meaning that if we make limit_table some how take a const reference to input, keeping input in scope because it stays in the inputs vector, we could actually still be able to recover from any error except for output->addToCache()

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.cpp
williamBlazing: send_message return what when its done? It usually would just add `inputs` to the output message cache. Multiple things in that `try` move that input. So I think this is a step we cannot RETRY

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.cpp
```diff
@@ -239,81 +261,86 @@ MergeAggregateKernel::MergeAggregateKernel(std::size_t kernel_id, const std::str
...
-    } else {
-        columns = ral::operators::compute_aggregations_with_groupby(
-                concatenated->toBlazingTableView(), mod_aggregation_input_expressions, mod_aggregation_types,
-                mod_aggregation_column_assigned_aliases, mod_group_column_indices);
+        eventTimer.stop();
+
+        auto log_output_num_rows = columns->num_rows();
+        auto log_output_num_bytes = columns->sizeInBytes();
+
+        // events_logger->info("{ral_id}|{query_id}|{kernel_id}|{input_num_rows}|{input_num_bytes}|{output_num_rows}|{output_num_bytes}|{event_type}|{timestamp_begin}|{timestamp_end}",
```
williamBlazing, line 325: why was this commented out? should we just delete it?

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.cpp
```diff
@@ -369,56 +369,74 @@ std::unique_ptr<ral::frame::BlazingTable> PartwiseJoin::join_set(
...
+    }catch(std::exception e){
+        return {ral::execution::task_status::FAIL, std::string(e.what()), std::move(inputs)};   
+    }
+
+	try{
+		this->leftArrayCache->put(std::stoi(args.at("left_idx")), std::move(left_batch));
+		this->rightArrayCache->put(std::stoi(args.at("right_idx")), std::move(right_batch));
+    }catch(rmm::bad_alloc e){
+        //can still recover if the input was not a GPUCacheData 
+        return {ral::execution::task_status::RETRY, std::string(e.what()), std::vector< std::unique_ptr<ral::frame::BlazingTable> > ()};
```
williamBlazing, line 432: If we dont have the source data anymore, we cant RETRY

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.cpp
```diff
@@ -938,89 +956,108 @@ void JoinPartitionKernel::small_table_scatter_distribution(std::unique_ptr<ral::
...
-			logger->error("{query_id}|{step}|{substep}|{info}|{duration}||||",
-										"query_id"_a=context->getContextToken(),
-										"step"_a=context->getQueryStep(),
-										"substep"_a=context->getQuerySubstep(),
-										"info"_a="In JoinPartitionKernel::do_process Invalid operation_type: {}"_format(operation_type),
-										"duration"_a="");
+			return {ral::execution::task_status::FAIL, std::string("In JoinPartitionKernel::do_process Invalid operation_type"), std::move(inputs)};
 		}
+	}catch(rmm::bad_alloc e){
+		return {ral::execution::task_status::RETRY, std::string(e.what()), input_consumed ? std::vector< std::unique_ptr<ral::frame::BlazingTable> > () : std::move(inputs)};
```
williamBlazing, line 1056: If we dont have the data, we cant RETRY

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.cpp
```diff
@@ -938,89 +956,108 @@ void JoinPartitionKernel::small_table_scatter_distribution(std::unique_ptr<ral::
...
+			std::string small_output_cache_name = scatter_left_right.first ? "output_a" : "output_b";
+			int small_table_idx = scatter_left_right.first ? LEFT_TABLE_IDX : RIGHT_TABLE_IDX;
+
+			broadcast(std::move(input),
+				this->output_.get_cache(small_output_cache_name).get(),
+				"", //message_id_prefix
+				small_output_cache_name, //cache_id
+				small_table_idx //message_tracker_idx
+			);
+		}	else if (operation_type == "big_table_passthrough") {
```
williamBlazing, line 977: The big table passthrough stuff, i think is no longer done in `do_process`. This looks like there was a bad merge somewhere. I think this happens in JoinPartitionKernel::run

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.cpp
williamBlazing: So, i do think we want to have `ral::operators::limit_table` take a const reference, so that we still have control of input. But `ral::operators::limit_table` may also have wanted to return `input` untouched. So i think we would want to do something like this:

```
std::unique_ptr<ral::frame::BlazingTable> limited_input;
bool output_is_just_input;
std::tie(limited_input, output_is_just_input, rows_limit) = ral::operators::limit_table(input, rows_limit);
```
Where if you have output_is_just_input == true, then limited_input is nullptr and you are supposed to just use input.

Then you also move `output->addToCache(std::move(input));` to outside the try catch, so that you can move the input into the RETRY.
And also, it would be something like this:

```
if (output_is_just_input)
     output->addToCache(std::move(input));
else
     output->addToCache(std::move(limited_input));
````

File: engine/src/execution_graph/logic_controllers/BatchProcessing.cpp [hunk 7]
```diff
@@ -156,10 +156,18 @@ TableScan::TableScan(std::size_t kernel_id, const std::string & queryString, std
...
-void TableScan::do_process(std::vector< std::unique_ptr<ral::frame::BlazingTable> > inputs,
+ral::execution::task_result TableScan::do_process(std::vector< std::unique_ptr<ral::frame::BlazingTable> > inputs,
     std::shared_ptr<ral::cache::CacheMachine> output,
     cudaStream_t /*stream*/, const std::map<std::string, std::string>& /*args*/) {
-    output->addToCache(std::move(inputs[0]));
+    try{
+        output->addToCache(std::move(inputs[0]));
+    }catch(rmm::bad_alloc e){
+        //can still recover if the input was not a GPUCacheData 
+        return {ral::execution::task_status::RETRY, std::string(e.what()), std::vector< std::unique_ptr<ral::frame::BlazingTable> > ()};
```
williamBlazing, line 166: cant have a retry without capturing the inputs

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
```diff
@@ -243,19 +251,38 @@ BindableTableScan::BindableTableScan(std::size_t kernel_id, const std::string &
...
+
+    try{
+        if(this->filtered) {
+            output->addToCache(std::move(columns));
+        } else {
+            output->addToCache(std::move(input));
+        }
+    }catch(rmm::bad_alloc e){
+        //can still recover if the input was not a GPUCacheData 
+        return {ral::execution::task_status::RETRY, std::string(e.what()), std::vector< std::unique_ptr<ral::frame::BlazingTable> > ()};
```
williamBlazing, line 281: cant have a retry without capturing the inputs

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
```diff
@@ -311,8 +337,12 @@ kstatus BindableTableScan::run() {
 
         std::unique_lock<std::mutex> lock(kernel_mutex);
         kernel_cv.wait(lock,[this]{
-            return this->tasks.empty();
+            return this->tasks.empty() || ral::execution::executor::get_instance()->has_exception();
```
williamBlazing, line 340: @rommelDB I am guessing that this is the prototype of how you want to handle the exceptions? and then you would add it to all the other kernels? If so, this PR should be WIP not in review

williamBlazing: I am wondering how this logic would behave if multiple exceptions occur.  We could end up only serving up one of the several exceptions. And additionally all the kernels would throw the same exception. How does this look to the user? DOes the user end up with multiple repeated exception messages or just one? @rommelDB could you please test out what would happen if multiple kernels throw an error at the same time

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
```diff
@@ -345,12 +375,21 @@ Projection::Projection(std::size_t kernel_id, const std::string & queryString, s
...
-    auto columns = ral::processor::process_project(std::move(input), expression, this->context.get());
-    output->addToCache(std::move(columns));
+
+    try{
+        auto & input = inputs[0];
+        auto columns = ral::processor::process_project(std::move(input), expression, this->context.get());
+        output->addToCache(std::move(columns));
+    }catch(rmm::bad_alloc e){
+        //can still recover if the input was not a GPUCacheData 
+        return {ral::execution::task_status::RETRY, std::string(e.what()), std::vector< std::unique_ptr<ral::frame::BlazingTable> > ()};
```
williamBlazing, line 388: You cant retry without capturing the inputs

engine/src/execution_graph/logic_controllers/BatchUnionProcessing.cpp
```diff
@@ -14,14 +14,29 @@ UnionKernel::UnionKernel(std::size_t kernel_id, const std::string & queryString,
...
+        return {ral::execution::task_status::RETRY, std::string(e.what()), std::move(inputs)};
+    }catch(std::exception e){
+        return {ral::execution::task_status::FAIL, std::string(e.what()), std::move(inputs)};   
+    }
+
+    try{
+        this->add_to_output_cache(std::move(input));
+    }catch(rmm::bad_alloc e){
+        //can still recover if the input was not a GPUCacheData 
+        return {ral::execution::task_status::RETRY, std::string(e.what()), std::vector< std::unique_ptr<ral::frame::BlazingTable> > ()};
```
williamBlazing, line 35: You cant RETRY without the inputs

williamBlazing (changes requested):
In general, i think we need to make sure that RETRYs have their data, unless i am mistaken.
The exception management logic, right now looks like a prototype in a couple places, we should finish that up here too

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
rommelDB: @williamBlazing Regarding the multiple exceptions, which would be the ideal output? at this time, when two kernels throw an exception, both error messages are shown to the user. Even if there are the same kernel or different ones.

rommelDB:
rerun tests

rommelDB:
rerun tests

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.cpp
rommelDB: I didn't find any reference to big_table_passthrough in the main branch, and this is the only reference to this operation_type, so it seems it is an incomplete feature. Should I remove it?

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.cpp
rommelDB: Fixed.

engine/src/execution_graph/logic_controllers/BatchAggregationProcessing.cpp
rommelDB: That code actually works fine. Uncommented.

engine/src/execution_graph/logic_controllers/BatchJoinProcessing.cpp
williamBlazing: Yes, we no longer need this. Please delete

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.cpp
```diff
@@ -525,39 +576,51 @@ LimitKernel::LimitKernel(std::size_t kernel_id, const std::string & queryString,
...
+            auto log_input_num_bytes = input->sizeInBytes();
+
+            std::unique_ptr<ral::frame::BlazingTable> limited_input;
+            bool output_is_just_input;
+
+            eventTimer.start();
+            std::tie(limited_input, output_is_just_input, rows_limit) = ral::operators::limit_table(input->toBlazingTableView(), rows_limit);
+            eventTimer.stop();
+
+            auto log_output_num_rows = limited_input->num_rows();
```
williamBlazing, line 598: here we want to check if output_is_just_input to know what we want to log

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
```diff
@@ -242,19 +254,38 @@ BindableTableScan::BindableTableScan(std::size_t kernel_id, const std::string &
...
 
-void BindableTableScan::do_process(std::vector< std::unique_ptr<ral::frame::BlazingTable> > inputs,
+ral::execution::task_result BindableTableScan::do_process(std::vector< std::unique_ptr<ral::frame::BlazingTable> > inputs,
     std::shared_ptr<ral::cache::CacheMachine> output,
     cudaStream_t /*stream*/, const std::map<std::string, std::string>& /*args*/) {
     auto & input = inputs[0];
-    if(this->filtered) {
-        auto columns = ral::processor::process_filter(input->toBlazingTableView(), expression, this->context.get());
-        columns->setNames(fix_column_aliases(columns->names(), expression));
+    std::unique_ptr<ral::frame::BlazingTable> columns;
```
williamBlazing, line 261: Can you rename std::unique_ptr<ral::frame::BlazingTable> columns;
to filtered_input, or something more descriptive than just columns. 
Because otherwise we have either input or columns, and from the names its hard to understand what is the difference

engine/src/execution_graph/logic_controllers/taskflow/executor.cpp
```diff
@@ -94,12 +115,44 @@ void task::run(cudaStream_t stream, executor * executor){
...
+    
+    if(task_result.status == ral::execution::task_status::SUCCESS){
+        complete();
+    }else if(task_result.status == ral::execution::task_status::RETRY){
+        std::size_t i = 0;
+        for(auto & input : inputs){
+            if  (input->get_type() == ral::cache::CacheDataType::GPU || input->get_type() == ral::cache::CacheDataType::GPU_METADATA){
+                //this was a gpu cachedata so now its not valid
+                if(input_gpu.size() > 0 && i <= input_gpu.size()){
+                    static_cast<ral::cache::GPUCacheData *>(input.get())->set_data(std::move(input_gpu[i]));
```
williamBlazing, line 129: input_gpu is no longer valid here because we moved it.
Here we want to try to recover the data from task_result

williamBlazing: We also want to make sure that do add a null check to the inputs we are getting back from task_result, because we are not going to be sure that we did not actually do a std::move on them during the execution of the task. If it fails the null check, then we should do whatever we do when the task_status is a FAIL

williamBlazing (changes requested):

PR #1259, opened 2020-12-11
[REVIEW] Feature/sortandsamplekernel executor

Christian8491:
This PR implements the `SortAndSampleKernel` as the executor model require.


GPUtester:

engine/src/distribution/primitives.cpp
```diff
@@ -29,13 +30,18 @@ typedef blazingdb::manager::Context Context;
 typedef blazingdb::transport::Node Node;
 
 
-
 std::unique_ptr<BlazingTable> generatePartitionPlans(
-				cudf::size_type number_partitions, const std::vector<BlazingTableView> & samples,
-				const std::vector<cudf::order> & sortOrderTypes) {
+	cudf::size_type number_partitions,
```
felipeblazing, line 34: why does this takea  table instead of a view? a table has to be moved a view cna be copied

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.cpp
```diff
@@ -86,16 +86,21 @@ SortAndSampleKernel::SortAndSampleKernel(std::size_t kernel_id, const std::strin
     this->query_graph = query_graph;
     set_number_of_message_trackers(2); //default
     this->output_.add_port("output_a", "output_b");
+    get_samples = true;
+    already_computed_partition_plan = false;
 }
 
-void SortAndSampleKernel::compute_partition_plan(std::vector<ral::frame::BlazingTableView> sampledTableViews, std::size_t avg_bytes_per_row, std::size_t local_total_num_rows) {
+void SortAndSampleKernel::compute_partition_plan(
+    std::vector<std::unique_ptr<ral::frame::BlazingTable>> inputSamples,
```
felipeblazing, line 94: why did we switch from views to tables again?

felipeblazing (changes requested):

engine/src/distribution/primitives.cpp
williamBlazing: The function that calls this one used to take views, but needed to now take tables. So we updated this one as well, since most of our code now operates on tables, not views

engine/src/execution_graph/logic_controllers/BatchOrderByProcessing.cpp
williamBlazing: This functions is now effectively a task, therefore it needs to operate on Tables not views

Issue #1177, opened 2020-11-18
[FEA] Support INITCAP for strings


beckernick:
I'd like to be able to use INITCAP to capitalize the first character of every "token" in a string. This is listed as a supported operation on string columns in the Calcite reference, but is currently not implemented.

```python
from pyspark.sql import SparkSession
from blazingsql import BlazingContext
import pandas as pd
​
​
# spark = SparkSession.builder \
#     .master("local") \
#     .getOrCreate()
​
# bc = BlazingContext()
​
​
df = pd.DataFrame({
    "a": ["this is", "an example", "of sql"],
    "b": [10, 9, 8],
})
​
bc.create_table("df", df)
sdf = spark.createDataFrame(df)
sdf.createOrReplaceTempView("df")
​
query = """
SELECT
    a,
    INITCAP(a)
FROM df
"""
​
spark.sql(query).show()
​
print(bc.explain(query))
# print(bc.sql(query)) # fails
+----------+----------+
|         a|initcap(a)|
+----------+----------+
|   this is|   This Is|
|an example|An Example|
|    of sql|    Of Sql|
+----------+----------+

LogicalProject(a=[$0], EXPR$1=[INITCAP($0)])
  BindableTableScan(table=[[main, df]], projects=[[0]], aliases=[[a, EXPR$1]])

```

derekmorr:
I'm happy to work on this if no one else is interested.

williamBlazing:
@derekmorr Welcome to the community! You are welcome to tackle this issue. I have assigned it to you. If you have any questions, please feel free to ask.

williamBlazing:
The implementation of this feature should look a little like this https://github.com/BlazingDB/blazingsql/pull/1193

derekmorr:
Just to clarify -- it looks like I have to first submit a PR to the tests repository with updated parquet files, then make a PR for this issue in this repo so that CI will run, is that correct?

williamBlazing:
Hello @derekmorr that is correct. I am working on making some of our development processes public to help people such as yourself to help out more easily, but I have been too busy. You can take a look at this which is from our internal documents:

Adding new End to End tests
 
To add a new End to End tests you have to follow the next steps:
If the new query is related to an existing script, you can add it at the end of the script inside:
$CONDA_PREFIX/blazingsql/test/BlazingSQLTest/EndToEndTests/existingScript.py
In case that the new query is part of a non-existent test suite, create a new script and add the new tests on the new script inside:
[...]

PR #1260, opened 2020-12-12
initcap support

derekmorr:
This addressed #1177


GPUtester:

williamBlazing:
add to whitelist

engine/src/parser/expression_utils.cpp
```diff
@@ -54,6 +54,7 @@ bool is_unary_operator(operator_type op) {
 	case operator_type::BLZ_CHAR_LENGTH:
 	case operator_type::BLZ_STR_LOWER:
 	case operator_type::BLZ_STR_UPPER:
+        case operator_type::BLZ_STR_INITCAP:
```
williamBlazing, line 57: please  fix spacing so that it matches the rest

engine/src/parser/expression_utils.cpp
```diff
@@ -135,6 +136,7 @@ cudf::type_id get_output_type(operator_type op, cudf::type_id input_left_type) {
 		return cudf::type_id::TIMESTAMP_NANOSECONDS;
 	case operator_type::BLZ_STR_LOWER:
 	case operator_type::BLZ_STR_UPPER:
+        case operator_type::BLZ_STR_INITCAP:
```
williamBlazing, line 139: please  fix spacing so that it matches the rest

engine/src/parser/expression_utils.cpp
```diff
@@ -285,6 +287,7 @@ operator_type map_to_operator_type(const std::string & operator_token) {
 		{"CHAR_LENGTH", operator_type::BLZ_CHAR_LENGTH},
 		{"LOWER", operator_type::BLZ_STR_LOWER},
 		{"UPPER", operator_type::BLZ_STR_UPPER},
+                {"INITCAP", operator_type::BLZ_STR_INITCAP},
```
williamBlazing, line 290: please  fix spacing so that it matches the rest

engine/src/parser/expression_utils.hpp
```diff
@@ -46,6 +46,7 @@ enum class operator_type {
 	BLZ_CHAR_LENGTH,
 	BLZ_STR_LOWER,
 	BLZ_STR_UPPER,
+        BLZ_STR_INITCAP,
```
williamBlazing, line 49: please  fix spacing so that it matches the rest

williamBlazing (changes requested):

williamBlazing:
Please update the CHANGELOG so that it triggers CI

engine/src/parser/expression_utils.hpp
derekmorr: addressed

engine/src/parser/expression_utils.cpp
derekmorr: addressed

engine/src/parser/expression_utils.cpp
derekmorr: addressed

engine/src/parser/expression_utils.cpp
derekmorr: addressed

derekmorr:
Changes should be addressed.

williamBlazing:
rerun-tests

williamBlazing:
It seems like a couple queries are failing to match against their results. Looking at the logs it almost seems like the results from drill are the ones that are incorrect. Its worth digging in more to see if that is the case. It would not be the first time we run into something like that

Christian8491:
Hi @derekmorr  could you please review the last two comments in order to fix these issues related to gpuCI jobs.

Issue #830, opened 2020-06-25
Stop depending on gtest for runtime


aucahuasi:
- [x] flag/pre compiler flag to NOT build debugingutils if we are releasing blazing for conda
- [x] remote the gtest dependency from our conda runtime meta yaml
- [x] fix the gtest version in our conda meta yaml for sections: build and host ... it should be the same version of cudf https://github.com/rapidsai/cudf/blob/branch-0.15/cpp/cmake/Templates/GoogleTest.CMakeLists.txt.cmake#L7

PR #1262, opened 2020-12-14
[REVIEW] Stop depending on gtest for runtime

aucahuasi:
https://github.com/BlazingDB/blazingsql/issues/830


GPUtester:

williamBlazing:
@aucahuasi How do you use this? If i want to build so that i can use the DebuggingUtils, what do I need to do in `./build.sh` ?
We probably want to expose that pre-compiler flag in `./build.sh`

aucahuasi:
> @aucahuasi How do you use this? If i want to build so that i can use the DebuggingUtils, what do I need to do in `./build.sh` ?
> We probably want to expose that pre-compiler flag in `./build.sh`

done https://github.com/aucahuasi/blazingsql/blob/feature/gtest-only-for-devs/engine/src/utilities/DebuggingUtils.h#L5

PR #1266, opened 2020-12-18
[REVIEW] Remove BlazingThread class

jeanp413:
Removing unnecessary BlazingThread class and use standard c++.
The std::future object will hold any exception and will throw it (if present) when calling the `get()` method


GPUtester:

PR #1267, opened 2020-12-18
[REVIEW] exception logging, orderby fixes and cache_machine with cache names

williamBlazing:
This PR does it all. Its the swiss army knife of PRs

- CacheMachine now has names and therefore improves logging so that you now know what is added to what cache
- Somethings needed to be refactored because of the above
- Fixed deadlocks in the executor
- Fixed deadlocks in the order by code
- Imrpved joins code by removing need to do a task for when there is a small table scatter, the big table passthrough does not need to be a task
- Improved logging in catch blocks and improved how some of those work
- Added retry logic to comms


GPUtester:

PR #1277, opened 2021-01-07
[REVIEW] Fix FileSystems paths when extension is not provided

### LINKED ISSUE #1276: Create table issue with GCP when extension is not provided
The following create table works fine.

`>>>bc.create_table('test_temp','gs://temp/gpu/scan_table/visit_dt=2020-06-24/op_cmpny_cd=WMT-K1/000006_0.orc',file_format='orc')`

But, if i rename the 000006_0.orc  to 000006_0  and create table.
`bc.create_table('test_temp1','gs://temp/gpu/scan_table/visit_dt=2020-06-24/op_cmpny_cd=WMT-K1/000006_0',file_format='orc')`

I run into the following error.
`**[performPartition]** error partitioning table.
[14:42:54.349] [error] |||In uri_data_provider::get_next. What: Path 'gs://temp/gpu/scan_table/visit_dt=2020-06-24/op_cmpny_cd=WMT-K1/000006_0' does not exist. File or directory paths are expected to be in one of the following formats: For local file paths: '/folder0/folder1/fileName.extension'    For local file paths with wildcard: '/folder0/folder1/*fileName*.*'    For local directory paths: '/folder0/folder1/'    For s3 file paths: 's3://registeredFileSystemName/folder0/folder1/fileName.extension'    For s3 file paths with wildcard: '/folder0/folder1/*fileName*.*'    For s3 directory paths: 's3://registeredFileSystemName/folder0/folder1/'    For gs file paths: 'gs://registeredFileSystemName/folder0/folder1/fileName.extension'    For gs file paths with wildcard: '/folder0/folder1/*fileName*.*'    For gs directory paths: 'gs://registeredFileSystemName/folder0/folder1/'    For HDFS file paths: 'hdfs://registeredFileSystemName/folder0/folder1/fileName.extension'    For HDFS file paths with wildcard: '/folder0/folder1/*fileName*.*'    For HDFS directory paths: 'hdfs://registeredFileSystemName/folder0/folder1/'|||||
Traceback (most recent call last):
[14:42:54.349] [error] |||In parseSchema. What: Path 'gs://temp/gpu/scan_table/visit_dt=2020-06-24/op_cmpny_cd=WMT-K1/000006_0' does not exist. File or directory paths are expected to be in one of the following formats: For local file paths: '/folder0/folder1/fileName.extension'    For local file paths with wildcard: '/folder0/folder1/*fileName*.*'    For local directory paths: '/folder0/folder1/'    For s3 file paths: 's3://registeredFileSystemName/folder0/folder1/fileName.extension'    For s3 file paths with wildcard: '/folder0/folder1/*fileName*.*'    For s3 directory paths: 's3://registeredFileSystemName/folder0/folder1/'    For gs file paths: 'gs://registeredFileSystemName/folder0/folder1/fileName.extension'    For gs file paths with wildcard: '/folder0/folder1/*fileName*.*'    For gs directory paths: 'gs://registeredFileSystemName/folder0/folder1/'    For HDFS file paths: 'hdfs://registeredFileSystemName/folder0/folder1/fileName.extension'    For HDFS file paths with wildcard: '/folder0/fold  File "<stdin>", line 1, in <module>
er1/*fileName*.*'    For HDFS directory paths: 'hdfs://registeredFileSystemName/folder0/folder1/'|||||
  File "/conda/envs/bsql/lib/python3.7/site-packages/pyblazing/apiv2/context.py", line 1949, in create_table
    local_files,
  File "/conda/envs/bsql/lib/python3.7/site-packages/pyblazing/apiv2/context.py", line 2254, in _parseSchema
    input, file_format_hint, kwargs, extra_columns, ignore_missing_paths
  File "bsql_engine/io/io.pyx", line 199, in cio.parseSchemaCaller
  File "bsql_engine/io/io.pyx", line 219, in cio.parseSchemaCaller
  File "bsql_engine/io/io.pyx", line 95, in cio.parseSchemaPython
cio.ParseSchemaError: [ParseSchema Error] Path 'gs://temp/gpu/scan_table/visit_dt=2020-06-24/op_cmpny_cd=WMT-K1/000006_0' does not exist. File or directory paths are expected to be in one of the following formats: For local file paths: '/folder0/folder1/fileName.extension'    For local file paths with wildcard: '/folder0/folder1/*fileName*.*'    For local directory paths: '/folder0/folder1/'    For s3 file paths: 's3://registeredFileSystemName/folder0/folder1/fileName.extension'    For s3 file paths with wildcard: '/folder0/folder1/*fileName*.*'    For s3 directory paths: 's3://registeredFileSystemName/folder0/folder1/'    For gs file paths: 'gs://registeredFileSystemName/folder0/folder1/fileName.extension'    For gs file paths with wildcard: '/folder0/folder1/*fileName*.*'    For gs directory paths: 'gs://registeredFileSystemName/folder0/folder1/'    For HDFS file paths: 'hdfs://registeredFileSystemName/folder0/folder1/fileName.extension'    For HDFS file paths with wildcard: '/folder0/folder1/*fileName*.*'    For HDFS directory paths: 'hdfs://registeredFileSystemName/folder0/folder1/'`

Christian8491:
This PR closes  #1276

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,7 @@
 - #1204 Reenable json parser
 - #1241 Fixed cython exception handling
 - #1243 Fixed wrong CHAR regex replacing
-- #1277 Check file format for files when extension is not provided
-
+- #1277 Add extension for FileSystems when extension is not provided
 
 
 # BlazingSQL 0.17.0 (December 10, 2020)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@
 - #1241 Fixed cython exception handling
 - #1243 Fixed wrong CHAR regex replacing
 - #1275 Fixed issue in StringUtil::findAndReplaceAll when there are several matches
-- #1277 Add extension for FileSystems when extension is not provided
+- #1277 Support FileSystems (GS, S3) when extension of the files are not provided
 
 
 # BlazingSQL 0.17.0 (December 10, 2020)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - #1204 Reenable json parser
 - #1241 Fixed cython exception handling
 - #1243 Fixed wrong CHAR regex replacing
+- #1275 Fixed issue in StringUtil::findAndReplaceAll when there are several matches
 - #1277 Add extension for FileSystems when extension is not provided
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - #1204 Reenable json parser
 - #1241 Fixed cython exception handling
 - #1243 Fixed wrong CHAR regex replacing
+- #1277 Check file format for files when extension is not provided
 
 
 
--- a/engine/src/cython/engine.cpp
+++ b/engine/src/cython/engine.cpp
@@ -53,25 +53,19 @@ std::pair<std::vector<ral::io::data_loader>, std::vector<ral::io::Schema>> get_l
 			tableSchema.in_file,
 			tableSchema.row_groups_ids);
 
-		std::string file_format_hint;
 		std::shared_ptr<ral::io::data_parser> parser;
 		if(fileType == ral::io::DataType::PARQUET) {
 			parser = std::make_shared<ral::io::parquet_parser>();
-			file_format_hint = "parquet";
 		} else if(fileType == gdfFileType || fileType == daskFileType) {
 			parser = std::make_shared<ral::io::gdf_parser>();
 		} else if(fileType == ral::io::DataType::ORC) {
-			file_format_hint = "orc";
 			parser = std::make_shared<ral::io::orc_parser>(args_map);
 		} else if(fileType == ral::io::DataType::JSON) {
-			file_format_hint = "json";
 			parser = std::make_shared<ral::io::json_parser>(args_map);
 		} else if(fileType == ral::io::DataType::CSV) {
-			file_format_hint = "csv";
 			parser = std::make_shared<ral::io::csv_parser>(args_map);
 		} else if(fileType == ral::io::DataType::ARROW){
 	     	parser = std::make_shared<ral::io::arrow_parser>(tableSchema.arrow_table);
-			file_format_hint = "arrow";
 		}
 
 		std::shared_ptr<ral::io::data_provider> provider;
@@ -86,7 +80,7 @@ std::pair<std::vector<ral::io::data_loader>, std::vector<ral::io::Schema>> get_l
 			provider = std::make_shared<ral::io::gdf_data_provider>(tableSchema.blazingTableViews, uri_values[i]);
 		} else {
 			// is file (this includes the case where fileType is UNDEFINED too)
-			provider = std::make_shared<ral::io::uri_data_provider>(uris, uri_values[i], file_format_hint);
+			provider = std::make_shared<ral::io::uri_data_provider>(uris, uri_values[i]);
 		}
 		ral::io::data_loader loader(parser, provider);
 		input_loaders.push_back(loader);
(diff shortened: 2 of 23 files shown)


GPUtester:

Christian8491:
run tests

williamBlazing:
So, what i am seeing is that `getPathWithNormalizedFolderConvention` used to always force an assumption that if there is no extension, then it must be a folder, and therefore add a slash. I think we want to completely get rid of that function, at least this version on the engine side.

I think what we should do, is validate paths on the python side, when a user creates a table:
The logic would look something like this:
```
if file_format is not specified:
    if path has no extension and no slash at the end:
         throw error to the user saying that if you dont have an extension, you have to specify the file_format or if its a directory, it needs to end in a slash
    else
         continue with path as is
else
    if path has no extension and no slash at the end:
        we asume that its a path with not extension. continue with path as in
    else
        continue with path as is
```

Which basically it boils down to:
if file_format is not specified and path has no extension and no slash at the end:
    throw error to the user saying that if you dont have an extension, you have to specify the file_format or if its a directory, it needs to end in a slash

@roaramburu @drabastomek @felipeblazing What do you guys think of this user experience

williamBlazing:
@roaramburu @drabastomek @felipeblazing What do you guys think of this user experience

To summarize. The current issue is that we are effectively assuming in a couple places in the code that if you have a path that does not have an extension, it assumes that its actually a directory, and slaps a / at the end. This obviously does not work for files that dont have extensions.
The reason this is a bit of a problem, is that for some filesystems, like GCS or S3 or HDFS (not sure which one is the problematic one), we cant really know if a path is a file or directory. This is what i recall from a long time ago. Which is why we are making these assumptions instead of actually checking.

So there are two proposals. 
1. Make it so that :
if file_format is not specified and path has no extension and no slash at the end:
throw error to the user saying that if you dont have an extension, you have to specify the file_format or if its a directory, it needs to end in a slash

2. Invest some time to see if we can get all filesystems to detect if a path with no extension and no slash is actually a file or directory. If we can, then use that somehow.

drabastomek:
If we make these checks in Python as you suggest determining if something is a file or a directory is trivial and built-in into Python. See below:

```python
import os

print(f'extensionless file? : {os.path.isfile("extensionless")}')
print(f'extensionless dir? : {os.path.isdir("extensionless")}')
print(f'with_extension.txt file? : {os.path.isfile("with_extension.txt")}')
print(f'with_extension.txt dir? : {os.path.isdir("with_extension.txt")}')
print(f'folder file? : {os.path.isfile("folder")}')
print(f'folder dir? : {os.path.isdir("folder")}')
```

prints out

```python
extensionless file? : True
extensionless dir? : False
with_extension.txt file? : True
with_extension.txt dir? : False
folder file? : False
folder dir? : True
```

williamBlazing:
@drabastomek the problem is that for S3 of GCS filesystems, python cant interact with those filesystems, so what you propose does not really work for all filesystems

drabastomek:
Ah, my bad. This said:

1. s3fs (fsspec derivation) has `isfile` and `isdir` implemented for technically for S3 we could use a similar logic [https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem.isdir](https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem.isdir)
2. gcfs has these methods implemented as well: [https://gcsfs.readthedocs.io/en/latest/api.html](https://gcsfs.readthedocs.io/en/latest/api.html)

Do we, under the hood, create the s3fs or gcfs objects when registering the s3 or gcp?

Christian8491:
I think the `getPathWithNormalizedFolderConvention` function generated the issue. So, removing this as William suggested will fix the issue. But just to avoid calling an S3 or GS API when the file does not have extension and the `file_format` arg was not passed, we need to throw an exception before call the `_parseSchema` function :
```
if ( file_format_hint == "undefined"  and extension == ""  and input[0][-1] != "/"  ):
                raise Exception(
                    "ERROR: if your input file doesn't have an extension, you have to specify the `file_format`. Or if its a directory, it needs to end in a slash"
                )
```

drabastomek:
Yeah, I saw your fix. Have been wondering if we could test if the file is extension-less or a directory ourselves and throw an Exception only if file does not have extension rather than when it's pointing to a directory. I agree that this fix will work.

PR #1285, opened 2021-01-18
[REVIEW] Infer PROTOCOL when Dask client is passed

### LINKED ISSUE #1283: Set PROTOCOL automatically when BlazingContext is passed a Dask Client
As discussed in https://github.com/rapidsai/gpu-bdb/pull/163#discussion_r557533172 , it would be nice to set the PROTOCOL configuration automatically if the BlazingContext is passed an active Dask Client.

AN already running Dask cluster could be set up with UCX or TCP, and this may not be known by the BlazingSQL user at runtime in order to properly configure PROTOCOL. It would be valuable to have this happen automatically.

cc @Christian8491

Christian8491:
This PR closes #1283

### DIFF
--- a/engine/src/cython/initialize.cpp
+++ b/engine/src/cython/initialize.cpp
@@ -31,6 +31,7 @@
 #include <blazingdb/io/Library/Logging/CoutOutput.h>
 #include <blazingdb/io/Library/Logging/Logger.h>
 #include "blazingdb/io/Library/Logging/ServiceLogging.h"
+#include <blazingdb/io/Util/StringUtil.h>
 
 #include "communication/CommunicationData.h"
 
@@ -754,10 +755,9 @@ std::pair<std::pair<std::shared_ptr<CacheMachine>,std::shared_ptr<CacheMachine> 
 		std::map<std::string, comm::node> nodes_info_map;
 
 		comm::blazing_protocol protocol = comm::blazing_protocol::tcp;
-		if(config_options.find("PROTOCOL") != config_options.end()){
-			if(config_options["PROTOCOL"] == "UCX"){
-				protocol = comm::blazing_protocol::ucx;
-			}
+		std::string protocol_value = StringUtil::toLower(config_options["PROTOCOL"]);
+		if (protocol_value == "ucx"){
+			protocol = comm::blazing_protocol::ucx;
 		}
 		ucp_context_h ucp_context = nullptr;
 		ucp_worker_h self_worker = nullptr;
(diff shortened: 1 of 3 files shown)


GPUtester:

Christian8491:
rerun tests

Christian8491:
rerun tests

williamBlazing:
We want to be able to have BSQL using TCP and Dask with UCX and vice-versa. This is something we are actually going to be testing very soon to make sure it works, so we dont want it to always just use whatever dask is using. I suggest we have `PROTOCOL` default to "AUTO", so if the user does not explicitly set it, it will be set by whatever dask is using. BUT if a user goes out of their way to set it, it should use what the user set.

engine/src/cython/initialize.cpp
```diff
@@ -754,10 +754,8 @@ std::pair<std::pair<std::shared_ptr<CacheMachine>,std::shared_ptr<CacheMachine>
 		std::map<std::string, comm::node> nodes_info_map;
 
 		comm::blazing_protocol protocol = comm::blazing_protocol::tcp;
-		if(config_options.find("PROTOCOL") != config_options.end()){
-			if(config_options["PROTOCOL"] == "UCX"){
-				protocol = comm::blazing_protocol::ucx;
-			}
+		if (config_options["PROTOCOL"] == "ucx"){
```
williamBlazing, line 757: Could you please make it so that its capitalization agnostic when it checks this. So like apply to_lower() or whatever the c++ version of that is

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -1415,6 +1416,10 @@ def __init__(
             ).encode()
 
         if dask_client is not None:
+            self.config_options["PROTOCOL".encode()] = parse_address(
```
williamBlazing, line 1419: We want to be able to have BSQL using TCP and Dask with UCX and vice-versa. This is something we are actually going to be testing very soon to make sure it works, so we dont want it to always just use whatever dask is using. I suggest we have PROTOCOL default to "AUTO", so if the user does not explicitly set it, it will be set by whatever dask is using. BUT if a user goes out of their way to set it, it should use what the user set.

williamBlazing (changes requested):

williamBlazing:
@Christian8491 Could you please also add commentary description to the `def __init__` of BlazingCOntext that talks about all the config_options. We are missing `PROTOCOL` in there

Issue #1305, opened 2021-02-02
Revamping Loggers: use events_logger in all do_process


aucahuasi:
Parent https://github.com/BlazingDB/blazingsql/issues/1297

Issue #1297, opened 2021-01-25
Revamping Loggers


williamBlazing:
The following is a summary of all the logs we have, what they log and what are their deficiencies:

**input_comms/output_comms**
- This logs all inbound and outgoing messages. 
- Its generates a lot of logs which are useful for debugging communication issues. The issue of there being too many logs here is not a big problem, since we can now turn them off by default.

**queries_logger**
- This logs the following: ral_id|query_id|start_time|plan|query
- The plan is displayed as a JSON text where Relational algebra steps can have "children".  This makes it easier to consume programatically, but not visually
- A big problem with this logger is that the plan it logs is the logical plan, not the physical plan

**kernels_logger**
- This logs the following: ral_id|query_id|kernel_id|is_kernel|kernel_type
- The kernels logged are both compute kernels and cacheMachines.
- This log is not meant to be consumed alone
- It logs in the CacheMachine constructors and kernel constructor

**kernels_edges_logger**
- This logs the following: ral_id|query_id|source|sink
- Which describes the relationship between the elements in `kernels_logger`
- Its only used in `graph::add_edge`

**events_logger**
- This logs the following: ral_id|query_id|kernel_id|input_num_rows|input_num_bytes|output_num_rows|output_num_bytes|event_type|timestamp_begin|timestamp_end
- The biggest problem with this logger is that its only used in a few of the kernel `::do_process` functions. If we want to use this, it needs to be universally in all `::do_process` functions. 

**cache_events_logger**
- This logs the following: ral_id|query_id|source|sink|num_rows|num_bytes|event_type|timestamp_begin|timestamp_end
- Its used in:
     - `BatchSequence::next()` (and other similar next() functions)
     - `OutputKernel::run()`
     - `kernel::add_to_output_cache`
- The problem with this is that we no longer really use the `::next()` functions. 
- For this logger to work, we would need to properly log absolutely all CacheData movement on and off caches. Perhaps except the communication caches, since we have specific communication logs.

**batch_logger**
- This logs all sorts of events. Its structure is meant to be flexible, but that also makes it non-homogeneous, making harder to digest. 
- This logger lots some things that likely should be logged in separate logs (and sometimes already is)
- The info in this logger is made to be relatively easy to consume on its own
- The following are examples of some things that it logs that is a little redundant with other logs:
         - Queries and their relational algebra. This is redundant with `queries_logger`
         - Cache_events: when things are added and removed from caches.
[...]

PR #1308, opened 2021-02-03
[REVIEW] Improve engine loggers

aucahuasi:
related to https://github.com/BlazingDB/blazingsql/issues/1305 and https://github.com/BlazingDB/blazingsql/issues/1297


GPUtester:

williamBlazing (changes requested):
There is one thing to fix in the code.
Additionally you will need to add the new task_logger and remove the events_logger in  `initialize.cpp`.
In `initialize.cpp` we create the logger and we also set the header.
You will also then want to look in context.py how all the loggers are handled. There is a config_option related to enabling the loggers, there you can swap the new task_logger and the events_logger, and in bc.log, i think the logs are addressed there as well

PR #1313, opened 2021-02-04
[REVIEW] getting and using ORC metadata

### LINKED ISSUE #795: Get ORC Metadata and use with skipdata processing
We can get the metadata using this: 
https://github.com/apache/orc/blob/master/c%2B%2B/include/orc/Statistics.hh

Christian8491:
This PR closes #795


GPUtester:

engine/src/io/data_parser/OrcParser.cpp
```diff
@@ -68,5 +69,21 @@ void orc_parser::parse_schema(
 	}
 }
 
+std::unique_ptr<ral::frame::BlazingTable> orc_parser::get_metadata(std::vector<data_handle> handles,
+	std::vector<std::shared_ptr<arrow::io::RandomAccessFile>> files, int offset) {
+	std::vector<size_t> num_stripes(files.size());
+	std::vector<cudf::io::parsed_orc_statistics> statistics(handles.size());
+	for(size_t file_index = 0; file_index < handles.size(); file_index++) {
+		std::string file_path = handles[file_index].uri.toString(true);
+		statistics[file_index] = cudf::io::read_parsed_orc_statistics(cudf::io::source_info{file_path});
```
williamBlazing, line 78: Here we dont want to use file_path, because if its a file on s3 or hdfs, it wont work. You need to do the RandorAccessFile `files`

williamBlazing (changes requested):
found one issue, please address it

engine/src/io/data_parser/OrcParser.cpp
Christian8491: done in the _3e6af7d_  commit

williamBlazing (changes requested):
small changes

PR #1336, opened 2021-02-16
[WIP] Improvements for the cache API

aucahuasi:


GPUtester:

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
```diff
@@ -479,14 +480,25 @@ std::unique_ptr<ral::cache::CacheData>  CacheMachine::get_or_wait_CacheData(size
 }
 
 std::unique_ptr<ral::frame::BlazingTable> CacheMachine::pullFromCache() {
+    static int global_index = -1;
```
williamBlazing, line 483: This cant be static. This has to be a member of the CacheMachine. 
Also, the same paradigm has to be applied to `pullCacheData()`

engine/src/execution_graph/logic_controllers/CacheData.cpp
```diff
@@ -30,6 +30,20 @@ std::string randomString(std::size_t length) {
 	return random_string;
 }
 
+std::string MetadataDictionary::get_value(std::string key) {
+    if (!this->has_value(key)) {
```
williamBlazing, line 34: For this function, if the key does not exist, please return empty string. No need to log a warning

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
williamBlazing: Also for `pullUnorderedFromCache()` make it so that if is_array_access == true, then just use `pullFromCache()`

williamBlazing (changes requested):

PR #1347, opened 2021-02-22
[REVIEW] Fixing issue when reading orc metadata from DATE dtype

Christian8491:


GPUtester:

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -2349,9 +2351,10 @@ def create_table(self, table_name, input, **kwargs):
                 parsedMetadata = parseHiveMetadata(table, uri_values)
                 table.metadata = parsedMetadata
 
-            if (
-                parsedSchema["file_type"] == DataType.PARQUET
-                or parsedSchema["file_type"] == DataType.ORC
+            # TODO: if still reading ORC metadata has issues then we can skip
+            # using get_metadata argument equals to False
+            if (parsedSchema["file_type"] == DataType.PARQUET) or (
```
williamBlazing, line 2356: The `get_metadata` parameter should apply to both parquet and orc. So it should be more like this:
```
if (get_metadata and (parsedSchema["file_type"] == DataType.PARQUET or parsedSchema["file_type"] == DataType.ORC)):
```

williamBlazing (changes requested):

williamBlazing:
also please add that new parameter to the comments in the beginning of `def create_table`

PR #1359, opened 2021-02-25
[WIP] Avoid cuda issues when free pinned memory

aucahuasi:
Make buffer providers singleton to avoid cuda issues with pinned memory de-allocation


GPUtester:

PR #1360, opened 2021-02-26
Fix JDK install for Power9

jglaser:
Update to a more recent version of the IBM Java SDK version 11 (the previous version was not available for download anymore and the build script failed).

Allow users to directly specify the path to `libjvm.so` in the `JAVA_HOME` environment variable, since the internal directory structure of the SDK has changed. Maybe it would make sense to make this the new documented default way, and eliminate the convoluted logic here
https://github.com/BlazingDB/blazingsql/blob/20f2fa94411c6a8b3557b8cc18b72d9220a2b07f/pyblazing/pyblazing/apiv2/context.py#L75-L95


GPUtester:

PR #1361, opened 2021-03-01
[REVIEW] Supporting first_value & last_value for Window Function

Christian8491:


GPUtester:

Christian8491:
rerun tests

engine/src/execution_graph/logic_controllers/BatchWindowFunctionProcessing.cpp
```diff
@@ -32,48 +33,93 @@ ComputeWindowKernel::ComputeWindowKernel(std::size_t kernel_id, const std::strin
...
+            std::vector<cudf::order> sortOrderTypes(column_partitioned_indices.size(), cudf::order::ASCENDING);
+            std::unique_ptr<cudf::column> pivot_indexes = cudf::lower_bound(table_view_to_partition,
+                                                                            table_view_to_partition_wo_repetitions->view(),
+                                                                            sortOrderTypes,
+                                                                            null_orders);
+            // we want to get the differences between two contiguous pivots
+            std::vector<cudf::size_type> pivot_indexes_host = ral::utilities::column_to_vector<cudf::size_type>(pivot_indexes->view());
+            std::vector<cudf::size_type> offsets;
+            for (size_t i = 0; i < pivot_indexes_host.size() - 1; ++i) {
+                offsets.push_back(pivot_indexes_host[i + 1] - pivot_indexes_host[i]);
```
williamBlazing, line 98: If you have a large amount of partitions, this offsets vector could be large. This means this operation you are doing here can be slow. Please optimize this. 
See if you can calculate the difference on GPU. If you cant easily do it with cudf, you should be able to use thrust

williamBlazing: Actually a much better way to do this whole thing is to do a join, which would likely be faster.
For example, you do your group by on the partition keys with the first or last value agregation. Then instead of the lowebound/offsets/repeat thing. You can instead do a join from on the partition keys that have aggregated, to the partition keys that have not been aggregated. 
The thing to be careful there is if the output of the join will have the results out of order than what you want. So test that out first. And you can also see if there is a sort based join, as opposed to a hash based join. I think a sort based join would keep the order the way you want, since you would be sorting the join keys which would be the same as the partition keys which should be sorted already anyways

williamBlazing: And to clarify, by a join here, i mean just using cudf join operations, locally on the batch being processed, not a full bsql join

williamBlazing (changes requested):

tests/BlazingSQLTest/EndToEndTests/windowFunctionTest.py
```diff
@@ -971,6 +969,120 @@ def executionTest():
                 fileSchemaType,
             )
 
+            queryId = "TEST_38"
+            query = """select first_value(c_nationkey) over
```
williamBlazing, line 973: This test is a little trivial, because you are asking for the first value of c_nationkey, but you are partitioning by c_nationkey, therefore the first_value in that partition is the same as every value in the partition. Its an interesting edge case, but its not a good test for `first_value`

williamBlazing: Please add another similar test case, where you are getting the first value of something that you are not partitioning by

tests/BlazingSQLTest/EndToEndTests/windowFunctionTest.py
```diff
@@ -971,6 +969,120 @@ def executionTest():
...
+                queryType,
+                worder,
+                "",
+                acceptable_difference,
+                use_percentage,
+                fileSchemaType,
+            )
+
+            queryId = "TEST_39"
+            query = """select last_value(c_nationkey) over
```
williamBlazing, line 996: Same comment as above

engine/src/execution_graph/logic_controllers/BatchWindowFunctionProcessing.cpp
```diff
@@ -32,48 +35,95 @@ ComputeWindowKernel::ComputeWindowKernel(std::size_t kernel_id, const std::strin
...
+            std::vector<std::pair<cudf::size_type, cudf::size_type>> columns_in_common;
+
+            // we just want the key columns, not the values column (which is the last column)
+            std::vector<cudf::size_type> left_column_indices(left_table->num_columns() - 1);
+            std::iota(left_column_indices.begin(), left_column_indices.end(), 0);
+
+            std::vector<cudf::size_type> right_column_indices(partitioned_table_view.num_columns());
+            std::iota(right_column_indices.begin(), right_column_indices.end(), 0);
+
+            std::unique_ptr<cudf::table> join_table = cudf::left_join(
```
williamBlazing, line 97: why are you doing a left_join as opposed to an inner_join? is it because of nulls?
Both sides of join should have the same keys, so it seems like an inner_join would work. 

Also, if you are doing a left join, I would have expected to have the left side be `partitioned_table_view` and not `left_table`. `left_table` contains the group by keys, which would contain the first or last value per partition. You would want to add the corresponding first or last value to `partitioned_table_view` which has all the values. Honestly, i am not sure how much it matters, except for  perhaps the expected order of the output, and the order matters.

tests/BlazingSQLTest/EndToEndTests/windowFunctionTest.py
williamBlazing: Another issue with these test queries is that you have an order by at the end, which means that the order that was part of the windowing gets overridden by the order by at the end. Please take out the order by at the end to make sure the order we end up with at the end of the window function is correct

williamBlazing (changes requested):

PR #1362, opened 2021-03-01
[WIP] give useful message when file extension is not recognized

### LINKED ISSUE #1343: [BUG] BSQL should give a warning if the file(s) provided don't have a recognized extension
**Describe the bug**
If you create a table with a file whose extension is `.log` BSQL cant infer what datatype it is, so the user needs to specify `file_format`. Right now, BSQL only gives the user an error is the file has no extension at all and there is no `file_format` specified. It should also give an appropriate error for any extension that is not obvious

Christian8491:
This PR closes #1343


GPUtester:

Christian8491:
rerun tests

Christian8491:
rerun tests

Issue #1363, opened 2021-03-02
Finalize BlazingContext properly on exit


aucahuasi:
In the C++ side:
* If we want to finalize all the engine properly we need to implement first many parts: thread interruptions, free memory from incomplete threads, notify every ral node to shutdown the processing, etc.
* The threads/futures in c++ cannot be cancel by itself we need to program that logic: We should implement some sort of interruption logic for the kernel threads/futures and then clean its memory. For now we are using getExecuteGraphResult for the finalize caller and this will ensure to free all the memory

In the python side:
- We need to free the memory from all workers invoking the finalizer. We cannot just send the finalizeCaller on each worker in the BlazingContext.__del__ because python garbage collector already deleted some objects related with dask, so the tornado loop from dask doesn't exists there.

Issue #1290, opened 2021-01-20
Allow for concurrent queries form a single BlazingContext


williamBlazing:
Right now when you do:
`bc.sql()` execution on the python script halts until that function call returns when the function returns with the result of the query.
You used to be able to use the option `return_futures` but that feature is now obsolete due to https://github.com/BlazingDB/blazingsql/pull/1289

On the other hand https://github.com/BlazingDB/blazingsql/pull/1289 makes it easy to implement multiple concurrent queries.

This feature request is to propose an API and user experience for multiple concurrent queries support from a single BlazingContext.

The proposed API would be something as follows:
Proposed API A
```
query0 = 'SELECT * FROM my_table where columnA > 0'
query1 = 'SELECT * FROM my_table where columnB < 0'
token0 = bc.sql(query0, return_token=True)  
token1 = bc.sql(query1, return_token=True)
result0 = bc.fetch(token0)
result1 = bc.fetch(token1)
```

In this case `token0` and `token1` would be int32s which are actually just the queryId.
In this case `bc.fetch` would halt execution until the results are available.
We would also implement a function (which would be optional) that would look like this:
`done = bc.is_query_done(token0)`
which would return a boolean, simply indicating if the query is done.

Other ways we could do this are:
Proposed API B:
```
token0 = bc.async_sql(query)
is_done = bc.async_sql(token0, get_status=True)  #this is the optional is_query_done API
result = bc.async_sql(token0)  # here its the same API, but since we are passing in an int instead of a string we would know that we are getting the result
```

Proposed API C:
```
token0 = bc.sql(query, return_token=True)
is_done = bc.sql(token0, get_status=True)  #this is the optional is_query_done API
result = bc.sql(token0)  # here its the same API, but since we are passing in an int instead of a string we would know that we are getting the result
```

Feel free to propose other APIs. 

This internally, this would just use the APIs that are now part of https://github.com/BlazingDB/blazingsql/pull/1289, which allow us to start a query, check its status and get the results. Internally what happens is that when multiple queries are running at the same time is that each query has its own graph, and each graph is generating compute tasks. The compute tasks are then processed by the executor as resources allow. Right now the tasks would be processed FIFO (with a certain amount of parallelism depending on resources and configuration). Eventually we can set prioritization policies for which tasks get done first. For example tasks from the first query to start are given priority, or tasks which are most likely to reduce memory pressure are prioritized, etc...

felipeblazing:
```
[...]

PR #1364, opened 2021-03-02
[REVIEW] Implement the concurrent API (sql with tokens, status, fetch)

aucahuasi:
Issue/feature request: https://github.com/BlazingDB/blazingsql/issues/1290

During the implementation we discover some issues that we'll address in the future https://github.com/BlazingDB/blazingsql/issues/1363


GPUtester:

tests/BlazingSQLTest/EndToEndTests/concurrentTest.py
```diff
@@ -0,0 +1,180 @@
...
+    data_types = [
+        DataType.DASK_CUDF,
+        DataType.CSV,
+        DataType.PARQUET,
+    ]  # TODO json
+
+    if Settings.execution_mode == ExecutionMode.GENERATOR:
+        data_types = [DataType.CSV]
+
+    create_tables(bc, data_types, dask_client, nRals, dir_data_lc, tables)
```
williamBlazing, line 130: How does this work? It seems like here you are creating all the tables for all the different data source types, and then running the queries. But in the queries, the table name is the same, so you are not running the queries on the different datasource types. Am i missing something here?

williamBlazing (changes requested):

tests/BlazingSQLTest/EndToEndTests/concurrentTest.py
aucahuasi: the last update fixed this!

Issue #1385, opened 2021-03-10
[BUG] Read ORC fails to read with metadata


mlahir1:
Read ORC fails to read with metadata. Also, with the addition of read_metadata argument, if i set this parameter to false. create_table works fine.

```
>>> bc.create_table('table','/root/data/*',file_format='orc')

/conda/envs/bsql/lib/python3.7/site-packages/distributed/utils.py in sync(loop, func, callback_timeout, *args, **kwargs)
    338     if error[0]:
    339         typ, exc, tb = error[0]
--> 340         raise exc.with_traceback(tb)
    341     else:
    342         return result[0]
/conda/envs/bsql/lib/python3.7/site-packages/distributed/utils.py in f()
    322             if callback_timeout is not None:
    323                 future = asyncio.wait_for(future, callback_timeout)
--> 324             result[0] = yield future
    325         except Exception as exc:
    326             error[0] = sys.exc_info()
/conda/envs/bsql/lib/python3.7/site-packages/tornado/gen.py in run(self)
    760 
    761                     try:
--> 762                         value = future.result()
    763                     except Exception:
    764                         exc_info = sys.exc_info()
/conda/envs/bsql/lib/python3.7/site-packages/distributed/client.py in _gather(self, futures, errors, direct, local_worker)
   1827                             exc = CancelledError(key)
   1828                         else:
-> 1829                             raise exception.with_traceback(traceback)
   1830                         raise exc
   1831                     if errors == "skip":
/conda/envs/bsql/lib/python3.7/site-packages/bsql_engine/io/io.pyx in cio.parseMetadataCaller()
    403 
    404 
--> 405 cpdef parseMetadataCaller(fileList, offset, schema, file_format_hint, args):
    406     cdef vector[string] files
    407     for file in fileList:
/conda/envs/bsql/lib/python3.7/site-packages/bsql_engine/io/io.pyx in cio.parseMetadataCaller()
    424       arg_values.push_back(str.encode(str(value)))
    425 
--> 426     resultSet = blaz_move(parseMetadataPython(files, offset, cpp_schema, str.encode(file_format_hint), arg_keys,arg_values))
    427 
    428     names = dereference(resultSet).names
/conda/envs/bsql/lib/python3.7/site-packages/bsql_engine/io/io.pyx in cio.parseMetadataPython()
[...]

PR #1387, opened 2021-03-10
[WIP] Skip getting orc metadata for decimal type

Christian8491:
This PR is a temporary work around for #1385 but it will skip when dtype is `DECIMAL`. If the issue is not due to a `DECIMAL` dtype then it will throw a more descriptive message about the dtype that is causing the issue,


GPUtester:

Issue #1383, opened 2021-03-09
[BUG] LEFT OUTER JOIN Drops Rows


mlahir1:
LEFT OUTER JOIN Drops Rows

```
import cudf
df = cudf.DataFrame()
df['a'] = range(5)
df['b'] = 1
bc.create_table('t', df)
bc.sql('''
Select * from t t1
LEFT OUTER JOIN t t2
    on (t1.a < 3) and t1.a = t2.a
''')
```

Output:
```
    a	b	a0	b0
0	0	1	0	1
1	1	1	1	1
2	2	1	2	1
```

Expected Output:
```
	a	b	a0	b0
0	0	1	0	1
1	1	1	1	1
2	2	1	2	1
3	3	1	3	
4	4	1	4	
```

williamBlazing:
Quick correction:
Expected Output:

	a	b	a0	b0
0	0	1	0	1
1	1	1	1	1
2	2	1	2	1
3	3	1		
4	4	1

williamBlazing:
This error is happening due to an error in implementation. 
In cudf there is currently no support for inequality joins. BlazingSQL gets around that by supporting inequality joins IF and only IF there is also an equality join ANDed to the inequality join as is in the case of this example `(t1.a < 3) and t1.a = t2.a`

The way BSQL handles this is by just doing a join using the equality `t1.a = t2.a` and then applying a filter on the results of the join `(t1.a < 3)`. This works well for inner joins, but as showcased here, does not work for outer joins.

PR #1390, opened 2021-03-10
[WIP] Fix for left outer join with inequality

williamBlazing:
This PR fixes Issue #1383


GPUtester:

williamBlazing:
This PR tried to fix left outer joins with inequalities but failed. The algorithm used here does not work in certain situations. In particular, if you have a left outer join with an equiality and an inequality, and one row from left table joins to two rows on the right table, but the inequality only joins to one row on the right table, then this algorithm does not work

Issue #1330, opened 2021-02-13
[BUG] Substring query on nulls give an error


rommelDB:
**Describe the bug**
The following substringTest query produces an error when the data contains nulls.

```
queryId = "TEST_09"
query = """select substring(c_comment, 5 , CHAR_LENGTH(c_comment) - 5 ),
            substring(c_comment, CHAR_LENGTH(c_comment) - 3 , CHAR_LENGTH(c_comment) - 1 ),
            substring(substring(c_comment, 3 , CHAR_LENGTH(c_comment) - 1 ), 1, 1 )
            from customer where c_custkey < 100"""
```
Current log:
```
=============== New query: TEST_09 - gdf =================
Steps Complete 0/2|       [22:25:01.659] [error] |||ERROR in kernel::process trying to do do_process. Kernel name is: Projection Kernel id is: 0|||||
[22:25:01.659] [error] |||ERROR in graph::execute. What: std::exception|||||
                          [22:25:01.672] [error] 2124329382|||In get_execute_graph_results. What: std::exception|||||     
>>>>>>>>  [RunExecuteGraph Error] std::exception
```
This is a simpler reproducer:
```
query = """select substring(c_comment, 5, CHAR_LENGTH(c_comment) - 5) from customer where c_custkey < 100"""
```
**Expected behavior**
The above queries should generate a right output.

**Environment overview**
 - Environment location: Bare-metal
 - Method of cuDF install: conda nightly branch-0.19
 - Method of BSQL install: From source, nightly branch-0.19

rommelDB:
Fixed by #1392

PR #1392, opened 2021-03-11
[REVIEW] Fix substrings with nulls

aucahuasi:
https://github.com/BlazingDB/blazingsql/issues/1330


GPUtester:

engine/src/execution_graph/logic_controllers/LogicalProject.cpp
```diff
@@ -289,7 +290,12 @@ std::unique_ptr<cudf::column> evaluate_string_functions(const cudf::table_view &
                 computed_end_column = cudf::make_column_from_scalar(*end_scalar, table.num_rows());
                 end_column = computed_end_column->view();
             }
-
+            cudf::numeric_scalar<int32_t> start_zero(0);
```
williamBlazing, line 293: Before you do this operation, please check to see if there are any nulls you need to replace. Otherwise anytime there are no nulls, then this is an extra operation and allocation you dont need to do

williamBlazing (changes requested):

engine/src/execution_graph/logic_controllers/LogicalProject.cpp
aucahuasi: good catch ty!

aucahuasi: done

PR #1394, opened 2021-03-11
[REVIEW] Disabled support for outer joins with inequalities

williamBlazing:
This PR fixes Issue #1383

Note that this PR first tried to actually fix the issue https://github.com/BlazingDB/blazingsql/pull/1390
But the algorithm does not work in all cases, so we had to disable the support we thought we had


GPUtester:

Issue #1395, opened 2021-03-11
Implement a prototype to a create table from other RDBMS


aucahuasi:
**Proposal 1**
Direct approach using only create table semantic:
```python
bc.create_table("dept_emp", "mysql://lucho:admin@localhost:3306/sampledb/dept_emp")
bc.create_table("titles", "postgres://luis:12345@localhost:3306/testdb/subchema/titles")
```

**Proposal 2**
2 step approach, similar to what we have with storage registration:
```python
bc.mysql("mysqldb1", "mysql://lucho:admin@localhost:3306/sampledb")
bc.create_table("dept_emp", "mysqldb1/dept_emp")

bc.postgres("pgdb1", "postgres://luis:12345@localhost:3306/testdb")
bc.create_table("titles", "pgdb1/subchema/titles")
```

cc @williamBlazing @felipeblazing @rommelDB

aucahuasi:
@drabastomek @roaramburu feel free to make any suggestions here

williamBlazing:
I think I like your Proposal 1 better.
The storage registration wants to act more like something that has a path. While for creating a table from other databases, there it can depend a lot on that database's APIs. In the case above, the communication with mysql or postgres looks like a path, but for other databases it might be more like:
`bc.create_table('table_name',from_database='google_big_query', schema=...,database=...,table=...,credentials=...)`

roaramburu:
I agree with William's comments. Proposal 1 makes more sense to me, I can understand there being more complex fields to fill out in creating these tables off of databases and therefore more arguments might be necessary.

Love it though!

aucahuasi:
@williamBlazing @roaramburu  thank you!
Yes if other RDBMS needs additional attributes to define the connection then we can ask for this arguments in the create_table method.
Maybe instead to ask a connection string for mysql/postgress we want to start asking the arguments for these too, so any registration will have the same semantics

aucahuasi:
Also, another idea, if the users want to filter and select specific columns we may offer the argument `columns = ["col1", "col2"]`

drabastomek:
I do like the proposal 1 as well. My only worry is with 'free flow' text for passwords but RDBMS were never good with that...

aucahuasi:
thanks @drabastomek agree, but for that kind of API the user relies on environment variables or other mechanism and from those they load the password inside the python variable to be used as argument for the create table.
[...]

PR #1396, opened 2021-03-11
[REVIEW] Create tables from other RDBMS

aucahuasi:
https://github.com/BlazingDB/blazingsql/issues/1395


roaramburu:
What does everyone think of the argument `from_sql`?

@drabastomek what do you think? I think `engine` might be better and more informative.

cc @felipeblazing @williamBlazing

drabastomek:
@aucahuasi @roaramburu 

I think `from_sql` might be a bit misleading as someone might think this is ok:

```python
bc.from_sql(bc.sql('SELECT * FROM <foo>')
```

How about `from_db` or `from_db_table`?

engine/src/cython/engine.cpp
```diff
@@ -65,22 +77,39 @@ std::pair<std::vector<ral::io::data_loader>, std::vector<ral::io::Schema>> get_l
...
 		} else if(fileType == ral::io::DataType::ARROW){
 	     	parser = std::make_shared<ral::io::arrow_parser>(tableSchema.arrow_table);
-		}
+		} else if(fileType == ral::io::DataType::MYSQL) {
+#ifdef MYSQL_SUPPORT
+      parser = std::make_shared<ral::io::mysql_parser>();
+      auto sql = ral::io::getSqlInfo(args_map);
+      provider = std::make_shared<ral::io::mysql_data_provider>(sql);
+      isSqlProvider = true;
+#endif
```
williamBlazing, line 86: there should be an else to the `#ifdef MYSQL_SUPPORT` so that if its not compiled with that support and someone tries to make a table with that, that is outputs an appropriate error

engine/src/cython/engine.cpp
```diff
@@ -65,22 +77,39 @@ std::pair<std::vector<ral::io::data_loader>, std::vector<ral::io::Schema>> get_l
...
+    } else if(fileType == ral::io::DataType::SQLITE) {
+#ifdef SQLITE_SUPPORT
+  //		parser = std::make_shared<ral::io::sqlite_parser>();
+  //    auto sql = ral::io::getSqlInfo(args_map);
+  //    provider = std::make_shared<ral::io::sqlite_data_provider>(sql.schema,
+  //                                                               sql.table,
+  //                                                               sql.table_filter,                                                          sql.table_batch_size);
+  //    isSqlProvider = true;
+#endif
+    }
```
williamBlazing, line 96: why is postgressql not in this list?

williamBlazing (changes requested):

engine/src/cython/engine.cpp
felipeblazing: i think its still in progress for a seperate pr

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
```diff
@@ -123,6 +124,11 @@ kstatus TableScan::run() {
     std::vector<int> projections(schema.get_num_columns());
     std::iota(projections.begin(), projections.end(), 0);
 
+    if (provider->is_sql()) {
+      auto sql_provider = std::dynamic_pointer_cast<ral::io::abstractsql_data_provider>(provider);
```
felipeblazing, line 128: This should happen at construction since we can determine that the moment the provider is created so it is not needed to do this casting or set these column indices here. that should happen during construction

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
```diff
@@ -266,6 +272,11 @@ kstatus BindableTableScan::run() {
         std::iota(projections.begin(), projections.end(), 0);
     }
 
+    if (provider->is_sql()) {
+      auto sql_provider = std::dynamic_pointer_cast<ral::io::abstractsql_data_provider>(provider);
```
felipeblazing, line 276: same as comment right above, this should be set at construction

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
```diff
@@ -278,6 +289,11 @@ kstatus BindableTableScan::run() {
             //this will allow us to prevent from having too many open file handles by being
             //able to limit the number of file tasks
             auto handle = provider->get_next(true);
+            if (provider->is_sql()) {
```
felipeblazing, line 292: you should not be adding any special logic here. This part of the code needs to be completely ignorant of what kind of provider it is

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
felipeblazing: also and more importantly, the kernels shoudl have no idea of what kind of provider they have or dont have. They should only use the functions that are common to the provider base class. That is part of the intention of these classes. Yoiu shouldnt need to add anything to them

engine/src/io/data_parser/ArgsUtil.cpp
```diff
@@ -242,11 +249,53 @@ std::string getDataTypeName(DataType dataType) {
...
+	case DataType::SQLITE: return "sqlite"; break;
 	default: break;
 	}
 
 	return "undefined";
 }
 
+sql_info getSqlInfo(std::map<std::string, std::string> &args_map) {
+  // TODO percy william maybe we can move this constant as a bc.BlazingContext config opt
+  const size_t DETAULT_TABLE_BATCH_SIZE = 1000;
```
felipeblazing, line 263: That feels like a very small number depending on the latency and overhead of fetching lets say 10k or 100k rows vs fetching 1k 10 or 100 times

engine/src/cython/io.cpp
```diff
@@ -47,41 +66,65 @@ TableSchema parseSchema(std::vector<std::string> files,
...
-			}
+    if (isSqlProvider) {
+        ral::io::data_handle handle = provider->get_next(false);
+        parser->parse_schema(handle, schema);
+        if (schema.get_num_columns() > 0){
+          got_schema = true;
+        }
+    } else {
+      while (!got_schema && provider->has_next()){
+        ral::io::data_handle handle = provider->get_next();
```
williamBlazing, line 105: What if we make `data_handle` an interface instead of a struct where we are shoving more mutually exclusive stuff into it?

If we did that, we could replace `handle.file_handle != nullptr` with `handle.is_valid()` which would apply to any type of data_handle and make a lot of other things cleaner

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
```diff
@@ -123,6 +124,11 @@ kstatus TableScan::run() {
     std::vector<int> projections(schema.get_num_columns());
     std::iota(projections.begin(), projections.end(), 0);
 
+    if (provider->is_sql()) {
+      auto sql_provider = std::dynamic_pointer_cast<ral::io::abstractsql_data_provider>(provider);
+      sql_provider->set_column_indices(projections);
```
williamBlazing, line 129: this logic of setting the column indices that you want should be handled by the CacheDataIO. all the other data types handle it there.

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
williamBlazing, line 276: this logic of setting the column indices that you want should be handled by the CacheDataIO. all the other data types handle it there.

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
```diff
@@ -278,6 +289,11 @@ kstatus BindableTableScan::run() {
             //this will allow us to prevent from having too many open file handles by being
             //able to limit the number of file tasks
             auto handle = provider->get_next(true);
+            if (provider->is_sql()) {
+              if (handle.sql_handle.row_count == 0) {
```
williamBlazing, line 293: This should be handled by the `has_next()` function that was just called. If it returned true, then the row count here should never be 0. Is there a reason we are doing this?

engine/src/io/data_parser/sql/AbstractSQLParser.h
```diff
@@ -0,0 +1,43 @@
...
+
+#include "io/data_parser/DataParser.h"
+#include "arrow/io/interfaces.h"
+#include <memory>
+#include <vector>
+
+namespace ral {
+namespace io {
+
+class abstractsql_parser : public data_parser {
```
williamBlazing, line 17: What is the purpose of this class? It looks almost the same as the original data_parser interface?

williamBlazing: for that matter, it looks like its not used anywhere. Please delete it if its not being used

engine/src/io/data_parser/sql/MySQLParser.cpp
```diff
@@ -0,0 +1,598 @@
...
+        std::vector<int32_t> *tmp = new std::vector<int32_t>();
+        tmp->resize(total_rows);
+        host_cols[col] = tmp;
+      } break;
+      case cudf::type_id::INT64: {
+        std::vector<int64_t> *tmp = new std::vector<int64_t>();
+        tmp->resize(total_rows);
+        host_cols[col] = tmp;
+      } break;
+      case cudf::type_id::UINT8: {
```
williamBlazing, line 131: Hmmmm, have you tried Unsigned integers? BSQL does not support unsigned integers yet. Not sure what would happen if you tried using them here.  I would guess we should leave this part here as is, and in the schema parsing, make sure that it maps to a signed integer

williamBlazing (changes requested):

engine/src/io/data_parser/sql/MySQLParser.cpp
felipeblazing: what we did before was make the smller types larger. though now that i mention it this is just the allocation logic basically. What is the purpose of allocating a std vector then casting it to void? Wouldnt it be the same to just allocate the space?

engine/src/io/data_parser/sql/MySQLParser.cpp
```diff
@@ -0,0 +1,598 @@
...
+      case cudf::type_id::LIST: {} break;
+      case cudf::type_id::DECIMAL32: {} break;
+      case cudf::type_id::DECIMAL64: {} break;
+      case cudf::type_id::STRUCT: {} break;
+      case cudf::type_id::NUM_TYPE_IDS: {} break;
+    }
+  }
+
+  int row = 0;
+  while (res->next()) {
```
felipeblazing, line 193: can mysql c api only get one row at a time?

engine/src/io/data_parser/sql/MySQLParser.cpp
```diff
@@ -0,0 +1,598 @@
...
+        case cudf::type_id::TIMESTAMP_NANOSECONDS:
+        // TODO percy it seems we don't support this yet
+        // case cudf::type_id::DURATION_DAYS: {} break;
+        // case cudf::type_id::DURATION_SECONDS: {} break;
+        // case cudf::type_id::DURATION_MILLISECONDS: {} break;
+        // case cudf::type_id::DURATION_MICROSECONDS: {} break;
+        // case cudf::type_id::DURATION_NANOSECONDS: {} break;
+        // case cudf::type_id::DICTIONARY32: {} break;
+        case cudf::type_id::STRING: {
+          std::string real_data = res->getString(mysql_col).asStdString();
```
felipeblazing, line 320: it might be faster to accumulate a vector of strings and make a cudf string column once you have a batch full

engine/src/io/data_parser/sql/PostgreSQLParser.cpp
```diff
@@ -0,0 +1,549 @@
...
+        if (fsize < 0) {  // STRING, STRUCT, LIST, and similar cases
+          auto *string_col = new cudf_string_col();
+          string_col->offsets.reserve(resultNtuples + 1);
+          string_col->offsets.push_back(0);
+          return static_cast<void *>(string_col);
+        }
+        // primitives cases
+        const cudf::type_id cudf_type_id = cudf_types[projection_index];
+        switch (cudf_type_id) {
+        case cudf::type_id::INT8: {
```
felipeblazing, line 71: alot of this code seems to be shared with the mysqlparser perhaps the abstract_sql_parser should implement some methods :)

engine/src/io/data_parser/sql/PostgreSQLParser.cpp
```diff
@@ -0,0 +1,549 @@
...
+  }
+
+  cudf::io::table_with_metadata tableWithMetadata;
+  std::vector<std::unique_ptr<cudf::column>> cudf_columns;
+  cudf_columns.resize(static_cast<std::size_t>(resultNfields));
+  for (const std::size_t projection_index : column_indices) {
+    cudf::type_id cudf_type_id = cudf_types[projection_index];
+    switch (cudf_type_id) {
+    case cudf::type_id::INT8: {
+      std::vector<std::int8_t> *vector =
```
felipeblazing, line 286: this whole pattern with the void * feels fragile i wonder if just managing the memory and types is better

engine/src/io/data_parser/sql/sqlcommon.h
```diff
@@ -0,0 +1,42 @@
...
+                   const std::vector<cudf::bitmask_type> &null_mask) {
+  auto d_chars = cudf::detail::make_device_uvector_sync(host_col->chars);
+  auto d_offsets = cudf::detail::make_device_uvector_sync(host_col->offsets);
+  auto d_bitmask = cudf::detail::make_device_uvector_sync(null_mask);
+  return cudf::make_strings_column(d_chars, d_offsets, d_bitmask);
+}
+
+template <typename T>
+static inline std::unique_ptr<cudf::column>
+build_fixed_width_cudf_col(size_t total_rows,
```
felipeblazing, line 27: i think there are already cudf helpers for this and this doesnt seem to be related to sql

engine/src/cython/io.cpp
```diff
@@ -47,41 +66,65 @@ TableSchema parseSchema(std::vector<std::string> files,
...
 		bool open_file = false;
-		while (provider->has_next()){
-			std::vector<ral::io::data_handle> handles = provider->get_some(64, open_file);
-			for(auto handle : handles) {
-				schema.add_file(handle.uri.toString(true));
-			}
-		}
+    if (!isSqlProvider) {
+      while (provider->has_next()){
+        std::vector<ral::io::data_handle> handles = provider->get_some(64, open_file);
```
felipeblazing, line 122: @williamBlazing it seems to be used here the abstractsql_dat_provider provides this method so its not just decorative

felipeblazing (changes requested):

engine/src/io/data_parser/sql/MySQLParser.cpp
williamBlazing: I would recommend making `host_cols` a `std::vector<std::vector<char>>` so that you can just set the size of each `std::vector<char>` the size in bytes that it needs to hold the data and you don't need to do `new` which can lead to memory leaks if not handled correctly. You can then do a reinterpret_cast when setting or getting the values.

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
williamBlazing: @felipeblazing what do you think?

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
williamBlazing: Ok, I looked at the code and understand why you need to do this. What I would suggest is that `mysql_data_provider` be able to always hold a result set. For example:
It could have two extra members:
```
data_handle current_handle;
bool have_current_handle;  // initialized to false
```
then
```
bool mysql_data_provider::has_next() {
    if (!this->table_fetch_completed  && !have_current_handle){
         fetch_data();
   }
  return (this->table_fetch_completed == false);
}
```
`fetch_data()` would have the logic that is in `get_next()` that obtains and builds the data_handle.
`have_current_handle` would be set to true by `fetch_data` and set to false by `get_next`.

You can then take it to the next step and make it so that when you call `get_next()` after it returns the handle, it runs `fetch_data` on a separate thread so that it is already fetching the next batch while the rest of the engine does parsing or anything else. This async behaviour could be configurable

engine/src/io/data_parser/sql/MySQLParser.cpp
williamBlazing: With regards to the range checking, if we have a value outside the range, we would just have to throw an error. We cant change the type midway. But that is better than nothing. I think we will implement Unsigned types soon, so this will hopefully be a short term issue

engine/src/io/data_parser/sql/PostgreSQLParser.cpp
williamBlazing: i agree

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -3053,6 +3094,27 @@ def sql(
                         query_table.input = [query_table.input]
                     currentTableNodes.append(query_table)
 
+            elif (
```
williamBlazing, line 3097: what is actually happening here? this is just copied from when you have a file. How does the engine distribute pulling data from mySQL from different nodes?

williamBlazing (changes requested):

engine/src/io/data_parser/sql/MySQLParser.cpp
aucahuasi: @felipeblazing yes unfortunately mysql is using the jdbc approach and there is nothing like fetch_bacth in the jdbc standard, I didn't find anything like "fetch_batch" in the native C neither. The good news is that I think the underlying implementation of those drivers have a buffer of results and the client API is just returning the latest item that was fetched.

engine/src/io/data_parser/sql/PostgreSQLParser.cpp
aucahuasi: yep that was the reason of those abstract classes (common spaces), but I din't have enough time to move the code :p, I'll be move it there cc @gcca

engine/src/io/data_parser/sql/PostgreSQLParser.cpp
aucahuasi: totally agree, we already did that, our first approach was using typed vectors, but the casting was an issue, also the handling for strings is different.
Let me try again and see if we can get ride of void*

engine/src/io/data_parser/sql/sqlcommon.h
aucahuasi: we could not found the builder or a sample code, perhaps we can move this to CudfUtils somewhere ... do you know where I can found this cudf helpers (there must not be the ones in cudf::testing) cc @rommelDB

io/src/FileSystem/Uri.h
```diff
@@ -24,6 +24,11 @@ class Uri {
 		const std::string & authority,
 		const Path & path,
 		bool strict = false);  // convenience ctor, will use Uri(scheme, auth, Path, strict)
+  Uri(const std::string &scheme,
```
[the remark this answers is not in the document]
aucahuasi: To save a URI for the internal representation of the sql batch we fetched, something like
"mysql://db/table/part?columns" ... perhaps is not a good idea yet but is always a good thing to have URI for the resources we are handling (data_hanle.uri)

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -912,6 +917,15 @@ def kwargs_validation(kwargs, bc_api_str):
...
             "local_files",
             "get_metadata",
+            # SQL Engines arguments
+            "from_sql",
+            "hostname",
+            "port",
+            "username",
+            "password",
+            "schema",
+            "table_filter",
```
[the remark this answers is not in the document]
aucahuasi: Even when the predicate push down is ready I would vote to leave this "pre-filter" feature there, because it offers more granular ways to create a sql table. cc @drabastomek @roaramburu @felipeblazing

engine/src/io/data_parser/sql/MySQLParser.cpp
aucahuasi: @williamBlazing that was the first implementation of `host_cols` and if we do that we need to iterate all the rows again to construct the buffer for the cudf column. The way is implemented right now is pretty good because we save like 2 loops over all the rows and the data structs are exactly what the cudf ctors want cc @felipeblazing @gcca

williamBlazing: Perhaps I am not explaining myself. Right now you are holding the data in vectors of specific types that you are creating with `new`, but then you are type-erasing them and holding them as `void*` pointers in `host_cols`. My suggestion was that you replace the individual typed vectors you are creating with `new` and only use type erased vectors of type char that you dont have to use `new`. You can do a reinterpret_cast on those vectors of char to give them a type when you want to put a value or create a cudf from them. The purpose of this suggestion is so that you dont have to use `new` which can lead to memory leaks if not handled correctly

engine/src/io/data_parser/sql/PostgreSQLParser.cpp
williamBlazing: I think this refactoring can be done when the postgresql or sqlite implementations are done

io/src/FileSystem/Uri.h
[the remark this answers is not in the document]
williamBlazing: ok

engine/src/cython/engine.cpp
aucahuasi: Done

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
aucahuasi: I've removed is_sql from the data_provider interface, done

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
aucahuasi: done

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
aucahuasi: done

engine/src/io/data_parser/ArgsUtil.cpp
aucahuasi: now is 100k, done

engine/src/cython/engine.cpp
```diff
@@ -65,22 +77,39 @@ std::pair<std::vector<ral::io::data_loader>, std::vector<ral::io::Schema>> get_l
...
 		}
 
-		if(fileType == ral::io::DataType::CUDF || fileType == ral::io::DataType::DASK_CUDF) {
-			// is gdf
-			provider = std::make_shared<ral::io::gdf_data_provider>(tableSchema.blazingTableViews, uri_values[i]);
-		} else {
-			// is file (this includes the case where fileType is UNDEFINED too)
-			provider = std::make_shared<ral::io::uri_data_provider>(uris, uri_values[i]);
-		}
+    if (!isSqlProvider) {
```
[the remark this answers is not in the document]
aucahuasi: Is just a flag and it simplifies the if statements later, if we rely on fileType we will then need to do if fileType!=mysql && fileType!=postgre ...etc

engine/src/cython/io.cpp
aucahuasi: Yes the idea is to have a nice interface here, however I think that is not part of the scope of this PR. There are many ways create a better `data_handle` concept and to really be data_source agnostic, but we lack the concept of data_source per se in our engine. I would propose to take this issue later, not for this PR cc @felipeblazing @gcca

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
aucahuasi: Done, the logic is now in the kernel ctors and the kernel::run is agnostic to the data provider cc @felipeblazing @gcca

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
aucahuasi: Done, now CacheDataIO handle that part

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
aucahuasi: @williamBlazing It seems the `async` optimization would be a nice to have, I've already resolved the main issue regarding to the `is_sql`, the use of `next` and the use of `CacheDataIO`. How about we implement async data_providers in a next PR? cc @felipeblazing @gcca

engine/src/io/data_parser/sql/MySQLParser.cpp
aucahuasi: I think that is what we are doing, we create and accumulate in host and the once is ready we create the cudf columns, perhaps we need more info about what do you have in mind here cc @gcca

engine/src/io/data_parser/sql/PostgreSQLParser.cpp
aucahuasi: @felipeblazing @williamBlazing If we get ride of bool we would need to iterate all the rows later to have the references for cudf, I think `void*` is not ideal but is what it offers the best performance for now (i.e. it minimizes a lot the travels over all the rows, at least 2 times: one for getting the refs and the other for adapting the refs for the cudf column ctors) cc @gcca

engine/src/io/data_parser/sql/MySQLParser.cpp
aucahuasi: So the issue is about code style I guess, the `new` is being used but we are also using the `delete` to avoid any leaks.
The approach you are suggesting was the first implementation and it didn't work, because some types cannot be casted even using `reinterpred_cast`, this will fails for long double and strings.
If the core issue is about the use of `new` then note that we just need to make sure we are deleting correctly, and that is the case here. cc @felipeblazing @gcca

engine/src/io/data_parser/sql/AbstractSQLParser.h
aucahuasi: `abstractsql_parser` is a _thing_ now, it allow us to define some common flows for all the sql parsers

aucahuasi: done

engine/src/io/data_parser/sql/PostgreSQLParser.cpp
aucahuasi: Great idea!

engine/src/io/data_provider/sql/AbstractSQLDataProvider.h
```diff
@@ -0,0 +1,75 @@
...
+ */
+class abstractsql_data_provider : public data_provider {
+public:
+  abstractsql_data_provider(const sql_info &sql);
+
+  virtual ~abstractsql_data_provider();
+
+  /**
+	 * Tries to get up to num_files data_handles. We use this instead of a get_all() because if there are too many files, 
+	 * trying to get too many file handles will cause a crash. Using get_some() forces breaking up the process of getting file_handles.
```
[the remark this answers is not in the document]
aucahuasi: done

pyblazing/pyblazing/apiv2/context.py
aucahuasi: This is the only remain issue, we keep working on this one cc @gcca

aucahuasi: ok I just sent a push, that should fix this issue ... tomorrow morning we are going to run some test to make sure distribution is fine, good catch @williamBlazing thanks!

engine/src/io/data_parser/sql/MySQLParser.cpp
felipeblazing: hes suggestiong you use a std::vector<char> no new and delete needed but it has another issue. the <char> there has no guaranteed alignment so however you allocate that vector you need to make sure its aligned to the largest type you want to represent.

felipeblazing: I agree that the whole type erasing a vector of something only to reinterpret cast it later is more problematic than the need to manage the memory resource.. it should be at least 8 byte aligned on a 64 bit system if im not mistaken so until we have 128 bit integers it might not matter

aucahuasi: great thanks, yeah and we are taking care to manage the memory resource there in the abstract_sql_parser, soon in the next PR postgresql and sqlite will use that same code so there will be only a single place where we are using new and delete.

aucahuasi: One solution we can try in the future is to use std::shared_ptr<void> instead of a naked void*, we can use std::unique_ptr here but that would require mor code to create custom dtors (std::shared_ptr infers the type for us) 
However I think dealing with a better style/approach for this code will be subject for another PR, the way it is now do the job and doesn't leak

pyblazing/pyblazing/apiv2/context.py
aucahuasi: Distribution is working fine now, thanks!

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -3090,6 +3134,27 @@ def sql(
                         query_table.input = [query_table.input]
                     currentTableNodes.append(query_table)
 
+            elif (
+                query_table.fileType == DataType.MYSQL
+                or query_table.fileType == DataType.SQLITE
+                # or query_table.fileType == DataType.
+            ):
+                if query_table.has_metadata():
```
williamBlazing, line 3142: This block of code was just copied form elsewhere. It does work, but its misleading and confusing. I would recommend replacing it with the following, which would be more explicit about what its doing:

```
currentTableNodes = []
for i in range(0, numSlices):
                bt = BlazingTable(self.name, self.input, self.fileType, args=self.args)
                bt.column_names = self.column_names
                bt.column_types = self.column_types
                currentTableNodes.append(bt)
```

williamBlazing (changes requested):

engine/src/cython/io.cpp
```diff
@@ -47,41 +66,63 @@ TableSchema parseSchema(std::vector<std::string> files,
 		parser = std::make_shared<ral::io::json_parser>(args_map);
 	} else if(fileType == ral::io::DataType::CSV) {
 		parser = std::make_shared<ral::io::csv_parser>(args_map);
-	}
+	} else if(fileType == ral::io::DataType::MYSQL) {
+#ifdef MYSQL_SUPPORT
+		parser = std::make_shared<ral::io::mysql_parser>();
+    auto sql = ral::io::getSqlInfo(args_map);
+    provider = std::make_shared<ral::io::mysql_data_provider>(sql, 0, 0);
+#endif
```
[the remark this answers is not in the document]
aucahuasi: For now PR only supports MySQL, perhaps we can let @gcca to finish the upcoming PR and there we can have the kind of changes you are suggesting cc @felipeblazing

pyblazing/pyblazing/apiv2/context.py
aucahuasi: I would apply this changes in the up coming refactor project for pyblazing, as you said the code works but the style can improve, how about we push this suggestion for the next PRs cc @drabastomek

felipeblazing (approved):

PR #1419, opened 2021-03-23
[REVIEW] Feature/concat cache machine timeout

### LINKED ISSUE #1414: Concatenating Cache should have a maximum delay
Whenever we have a concatenating cache we don't want it to wait forever to feed forward when tables are very small so after a certain amount of time we should feed the information forward. I think something along the order of 100ms should be ok

williamBlazing:
Resolves #1414 

This PR introduces a time out to ConcatenatingCacheMachine, so that it will either wait to accumulate a certain number of bytes, or for a timeout (which is configurable) or for the CacheMachine to `finish()`.


GPUtester:

beckernick:
This resolves https://github.com/rapidsai/gpu-bdb/issues/201

PR #1422, opened 2021-03-26
[WIP] Fix for predict file type when directory is provided as input

### LINKED ISSUE #1420: [BUG] No longer offers file type from extension
**Describe the bug**
Unable to create table off a directory without specifying the file format, even if all files have the same and a descriptive extension, such as `.parquet`

60B8624A-1EF7-4543-BC03-8D9D76ABBE83

Christian8491:
This PR closes #1420

### DIFF
--- a/pyblazing/pyblazing/apiv2/context.py
+++ b/pyblazing/pyblazing/apiv2/context.py
@@ -2270,15 +2270,6 @@ class BlazingContext(object):
             # /path/to/data/folder/ -> name_file = /path/to/data/folder/, extension = ''
             name_file, extension = os.path.splitext(input[0])
 
-            if not recognized_extension(extension) and file_format_hint == "undefined":
-                raise Exception(
-                    "ERROR: Your input file doesn't have a recognized extension, "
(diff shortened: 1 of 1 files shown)


GPUtester:

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -2296,6 +2287,28 @@ def create_table(self, table_name, input, **kwargs):
                         kwargs["names"].pop(id)
                         kwargs["dtype"].pop(id)
 
+            # if the input is a directory and files do not have extension we want to raise an error
+            if name_file[-1] == "/" and extension == "":
+                all_files = os.listdir(name_file)
```
williamBlazing, line 2292: Something is wrong here. If we are creating a table from a folder, we will get the files listed out by the C++ layer. I cant remember where or how that happens, but the python layer cant be the one doing it. If you have an HDFS or S3 or GCS filesystem, python cant do `os.listdir`

williamBlazing (changes requested):

PR #1367, opened 2021-03-03
[REVIEW] OverlapAccumulatorKernel

williamBlazing:
This PR introduces a new kernel which will be used, in conjunction with other kernels yet to be implemented, to allow for window functions that do not have a partition by clause and have bounded frames

PR #1426, opened 2021-03-30
[REVIEW] Window Functions without partitioning

williamBlazing:
This PR Includes #1367 . I implements the ability to run window functions that do not have partitions but that do have bounded windows


GPUtester:

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
```diff
@@ -136,6 +136,10 @@ void CacheMachine::put(size_t index, std::unique_ptr<ral::frame::BlazingTable> t
 	this->addToCache(std::move(table), this->cache_machine_name + "_" + std::to_string(index), true);
 }
 
+void CacheMachine::put(size_t index, std::unique_ptr<ral::cache::CacheData> cacheData) {
+	this->addCacheData(std::move(cacheData), this->cache_machine_name + "_" + std::to_string(index), true);
```
felipeblazing, line 140: this doesnt feel quite right. Is there a reason we keep using these strings concatted to gether for naming  these portions.

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
```diff
@@ -686,6 +683,41 @@ bool CacheMachine::has_messages_now(std::vector<std::string> messages){
 	return true;
 }
 
+std::unique_ptr<ral::cache::CacheData> CacheMachine::pullAnyCacheData(const std::vector<std::string> & messages) {
```
felipeblazing, line 686: we are starting to accumulate way oto many ways of pulling data from these caches. I think its time we consolidate and have some kind of flag to handle all these different use cases

engine/src/execution_graph/logic_controllers/CacheMachine.h
```diff
@@ -51,13 +51,15 @@ class CacheMachine {
...
+	virtual void put(size_t index, std::unique_ptr<ral::cache::CacheData> cacheData);
+
 	virtual std::unique_ptr<ral::frame::BlazingTable> get_or_wait(size_t index);
 
 	virtual std::unique_ptr<ral::cache::CacheData> get_or_wait_CacheData(size_t index);
 
 	virtual void clear();
 
-	virtual bool addToCache(std::unique_ptr<ral::frame::BlazingTable> table, std::string message_id = "", bool always_add = false, const MetadataDictionary & metadata = {}, bool include_meta = false, bool use_pinned = false );
+	virtual bool addToCache(std::unique_ptr<ral::frame::BlazingTable> table, std::string message_id = "", bool always_add = false, const MetadataDictionary & metadata = {}, bool use_pinned = false );
```
felipeblazing, line 62: this adds yet parameter these apis are starting to be confusing

felipeblazing (changes requested):
This is a big set of changes but I dont see any unit tests made for all of the new primitives and functions that we added. I think we need to have at least some amount of coverage in the c++ testing layer

engine/src/execution_graph/logic_controllers/CacheMachine.h
williamBlazing: i removed a parameter to simplify, i did not add

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
williamBlazing: do you have any suggestions? i am not sure what is a better alternative

engine/src/execution_graph/logic_controllers/CacheMachine.cpp
williamBlazing: i am glad you think it does not feel quite right, its gross! but its how we currently can have an array access implementation on top of a queue. I have been wanting to create a new underlying data structure to hold the batches that more natively and permanently allow for both queue APIs and array APIs. OR having two different CacheMachines types that have different APIs. 
I figured we can look into making this better whenever we make more wholesale changes to CacheMachine and/or BlazingTable and/or CacheDAta to allow for multiple kernels to pull the same data from a CacheMachine

engine/src/execution_graph/logic_controllers/BatchWindowFunctionProcessing.cpp
```diff
@@ -214,38 +258,773 @@ ral::execution::task_result ComputeWindowKernel::do_process(std::vector< std::un
...
+    std::tie(this->preceding_value, this->following_value) = get_bounds_from_window_expression(this->expression);
+
+    ral::cache::cache_settings cache_machine_config;
+	cache_machine_config.type = ral::cache::CacheType::SIMPLE;
+	cache_machine_config.context = context->clone();
+    cache_machine_config.is_array_access = true;
+
+    std::string batches_cache_name = std::to_string(this->get_id()) + "_batches";
+    this->batches_cache = ral::cache::create_cache_machine(cache_machine_config, batches_cache_name);
+    std::string preceding_cache_name = std::to_string(this->get_id()) + "_preceding";
```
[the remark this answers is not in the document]
williamBlazing: I will see how i can make that better

engine/src/execution_graph/logic_controllers/BatchWindowFunctionProcessing.cpp
```diff
@@ -214,38 +258,773 @@ ral::execution::task_result ComputeWindowKernel::do_process(std::vector< std::un
...
+                    "duration"_a=timer.elapsed_time(),
+                    "kernel_id"_a=this->get_id());
+    }
+
+    return kstatus::proceed;
+}
+
+
+
+// START OverlapAccumulatorKernel
```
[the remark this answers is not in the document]
williamBlazing: will do

engine/src/execution_graph/logic_controllers/BatchWindowFunctionProcessing.cpp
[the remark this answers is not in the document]
williamBlazing: These caches are internal only to the kernel. They do not connect to other kernels. The names given to these caches here are only so that the logger for the cache machines has a unique name for it. They are not used in the logic anywhere, so I dont think that its necessary to use string constants

felipeblazing (approved):

PR #1444, opened 2021-04-07
added update to version for gcp

mario21ic:


GPUtester:

mario21ic:
rerun tests

romulo-auccapuclla:
rerun tests

Christian8491:
rerun tests

williamBlazing:
@mario21ic what is the dependencies.sh file? what is it meant to do? I notice that the cuda version and the python version are both hardcoded, which is concerning.

mario21ic:
> @mario21ic what is the dependencies.sh file? what is it meant to do? I notice that the cuda version and the python version are both hardcoded, which is concerning.

The file dependencies.sh was for avoiding forgetting some dependencies, but it could be a way to install our dependencies on the local environment. In a next PR I will add some arguments to be more flexible.

PR #1445, opened 2021-04-07
[REVIEW] Support for CURRENT_DATE, CURRENT_TIME and CURRENT_TIMESTAMP

### LINKED ISSUE #1377: [FEA] Implement CURRENT_DATE/TIME to pull the current time/date col
Implement functionality to add a new column with Current_date/time

**Describe the solution you'd like**
select CURRENT_DATE as date from table

**Describe alternatives you've considered**
can add more functionality around choosing the local time/various time zones

Christian8491:
This PR closes #1377 . Also in  #1377 we can find interesting out  discussion about this PR.


GPUtester:

Christian8491:
rerun tests

engine/src/CalciteExpressionParsing.cpp
```diff
@@ -134,9 +134,19 @@ std::unique_ptr<cudf::scalar> get_scalar_from_string(const std::string & scalar_
 	if(type.id() == cudf::type_id::TIMESTAMP_DAYS) {
 		return strings::str_to_timestamp_scalar(scalar_string, type, "%Y-%m-%d");
 	}
-	if(type.id() == cudf::type_id::TIMESTAMP_SECONDS || type.id() == cudf::type_id::TIMESTAMP_MILLISECONDS
-		|| type.id() == cudf::type_id::TIMESTAMP_MICROSECONDS || type.id() == cudf::type_id::TIMESTAMP_NANOSECONDS) {
-		if (scalar_string.find(":") != std::string::npos){
+	if(type.id() == cudf::type_id::TIMESTAMP_SECONDS) {
+		if (scalar_string.find("-") != std::string::npos && scalar_string.find(":") != std::string::npos){
```
williamBlazing, line 138: I see here that you are checking for `-` and `:`, but it might be better to use regex. This way we know for a fact that it will match the format described

engine/src/CalciteExpressionParsing.cpp
```diff
@@ -134,9 +134,19 @@ std::unique_ptr<cudf::scalar> get_scalar_from_string(const std::string & scalar_
...
+			return strings::str_to_timestamp_scalar(scalar_string, type, "%Y-%m-%d %H:%M:%S");
+		} else {
+			return strings::str_to_timestamp_scalar(scalar_string, type, "%Y-%m-%d");
```
williamBlazing, line 141: This else assumes a specific format, but we are not checking for that format. So you should add an if here, and then in the else throw an error if its not going to be a supported format.
You could also be more ambitious and add more formats, for example "%Y/%m/%d"

engine/src/CalciteExpressionParsing.cpp
```diff
@@ -134,9 +134,19 @@ std::unique_ptr<cudf::scalar> get_scalar_from_string(const std::string & scalar_
...
+		}
+	}
+	if(type.id() == cudf::type_id::TIMESTAMP_MILLISECONDS || type.id() == cudf::type_id::TIMESTAMP_MICROSECONDS
+		 || type.id() == cudf::type_id::TIMESTAMP_NANOSECONDS) {
+		if (scalar_string.find("-") != std::string::npos && scalar_string.find(":") != std::string::npos && scalar_string.find(".") != std::string::npos){
+			return strings::str_to_timestamp_scalar(scalar_string, type, "%Y-%m-%d %H:%M:%S.%f");
+		}
+		else if (scalar_string.find("-") != std::string::npos && scalar_string.find(":") != std::string::npos){
 			return strings::str_to_timestamp_scalar(scalar_string, type, "%Y-%m-%d %H:%M:%S");
 		} else {
```
williamBlazing, line 151: same as my comments above.

williamBlazing (changes requested):
Since you added/modified code related to interpreting literal timestamp or date values, could you please make sure we have E2E tests that have timestamp and date literals in different formats

felipeblazing (changes requested):
One question that I have  about this is are we comfortable that the timestamp could be different for different rows depending on when this invoked on each worker

felipeblazing:
> > One question that I have about this is are we comfortable that the timestamp could be different for different rows depending on when this invoked on each worker
> 
> Oh shit!!! I had not thought of that! Damn.....
> What if we always send a current time variable that comes from the python side via dask to the engine. It could come in handy in the future for other things as well. This time could be added to the context

I really dislike the notion of the python side being responsible for doing this as it feels like it should be part of the cpp side but I don't see too many other viable options since we don't want to add a synchronization step to projection kernels. It just feels weird that a runquery will need to take as an input the time it started.

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -1416,6 +1424,9 @@ def __init__(
                     It should use what the user set. If the user does not explicitly set it,
                     by default it will be set by whatever dask client is using ('tcp', 'ucx', ..).
                     NOTE: This parameter only works when used in the BlazingContext.
+            CURRENT_TIMESTAMP: A string representing the current timestamp in order to get the same
```
williamBlazing, line 1427: I dont think the CURRENT_TIMESTAMP should be a `config_option`. Its not something we want to have a user be able to set. Its also not really optional.

I think it should be something that is passed in like ctxToken, and similarly, like it, be added to the `Context`

williamBlazing (changes requested):

williamBlazing (approved):

felipeblazing (approved):

PR #1447, opened 2021-04-08
[REVIEW] Improve getting estimated output num rows

### LINKED ISSUE #1397: improve estimate of number of output rows for filter kernel
in our Filter Kernel we currentl estimate the number of output rows this way 

```  
std::pair<bool, uint64_t> Filter::get_estimated_output_num_rows(){
        std::pair<bool, uint64_t> total_in = this->query_graph->get_estimated_input_rows_to_kernel(this->kernel_id);
        if (total_in.first){
            double out_so_far = (double)this->output_.total_bytes_added();
            double in_so_far = (double)this->total_input_bytes_processed;
            if (in_so_far == 0){
                return std::make_pair(false, 0);
            } else {
                return std::make_pair(true, (uint64_t)( ((double)total_in.second) *out_so_far/in_so_far) );
            }
        } else {
            return std::make_pair(false, 0);
        }
    }
```

```
            double out_so_far = (double)this->output_.total_bytes_added();
            double in_so_far = (double)this->total_input_bytes_processed;
```
In particular is problematic. What we really want to know is the number of rows added to the output vs the number of input rows processed. We are interested in rows here more than bytes because not all columns are of fixed width (like strings).

Christian8491:
This PR closes #1397

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - #1331 Added flag to enable null e2e testing
 - #1418 Adding support for docker image
 - #1419 Added concat cache machine timeout 
+- #1447 Improve getting estimated output num rows
 
 ## Bug Fixes
 - #1335 Fixing uninitialized var in orc metadata and handling the parseMetadata exceptions properly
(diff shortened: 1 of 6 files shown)


GPUtester:

PR #1455, opened 2021-04-09
[REVIEW] Support for IS NOT FALSE condition

### LINKED ISSUE #1446: [BUG] IS NOT FALSE throws an RunExecuteGraph exception
**Describe the bug**
SQL call `SELECT * FROM <foo> WHERE <bool_column> IS NOT FALSE` prints out an Exception message from the graph
```python
>>>>>>>>  [RunExecuteGraph Error] std::exception
```
while `SELECT * FROM <foo> WHERE <bool_column> IS TRUE` or (conversely) `SELECT * FROM <foo> WHERE <bool_column> IS NOT TRUE` are fine.

Moreover, the code above prints an exception message in Jupyter but does not throw a proper Exception.

```python
import blazingsql as bsql
import cudf

df = cudf.DataFrame({'bools': [True, False, True, True, False]})
bc = bsql.BlazingContext()
bc.create_table('bools_table', df)

bc.sql('SELECT * FROM bools_table WHERE bools IS NOT FALSE')
```

**Expected behavior**
Should return all the rows where values is `TRUE`.

 - Environment location: Docker
 - Method of BlazingSQL install: Docker
   - `docker pull rapidsai/rapidsai:0.18-cuda11.0-runtime-ubuntu18.04-py3.8`
 - **BlazingSQL Version**: `'acce185359ed1b04a569e870436937bc74a4cf24'`

Christian8491:
This PR closes #1446 . This PR also enables `booleanTest` to use null values.

Some of these changes are related to fix 0.20 dependency conflicts and cuda compiler issues cc @mario21ic  @romulo-auccapuclla . 

As we know in order to support cuda version 11 (as discussed here https://github.com/rapidsai/rmm/issues/736)  we need to update our cuda driver versions (450) as  http://docs.nvidia.com/deploy/cuda-compatibility/index.html  suggest 

Finally it's expected that all the `cuda-10.X` jobs fails.


Christian8491:
rerun tests

Christian8491:
rerun tests

Christian8491:
rerun tests

Christian8491:
rerun tests

Christian8491:
rerun tests

williamBlazing:
rerun tests

Christian8491:
rerun tests

williamBlazing:
@Christian8491  `column_A IS NOT FALSE` seems to evaluate to just `column_A`, but, have you checked to make sure that that logic holds true with nulls?

Christian8491:
@williamBlazing  I will test using `nulls` and see what is the behavior.

Christian8491:
rerun tests

romulo-auccapuclla:
rerun tests

romulo-auccapuclla:
rerun

mario21ic:
rerun tests

romulo-auccapuclla:
rerun tests

mario21ic:
rerun tests

engine/src/Interpreter/interpreter_ops.cuh
```diff
@@ -664,11 +664,17 @@ private:
 			}
 			bool left_valid = getColumnValid(row_valids, left_position);
 
-			if(oper == operator_type::BLZ_IS_NULL) {
+			if (oper == operator_type::BLZ_IS_NOT_TRUE) {
+				left_valid = (left_value == false) ? true : false;
```
williamBlazing, line 668: This logic does not look right. Please verify.
Also, if this logic is not correct, how is this passing e2e tests?

williamBlazing: same questions about BLZ_IS_NOT_FALSE

williamBlazing: The way this code is written, it look like its either true or null, and never false. I think its actually either true or false and never null.

williamBlazing (changes requested):
Can you please also explain in the PR what do the changes to the Calcite side do?

engine/src/Interpreter/interpreter_ops.cuh
Christian8491: For a boolean column, `left_value` is also `boolean`. 
When the `value_i` is `null` or `false`, then `left_value` will be`False`.

Christian8491: The next column diagram illustrates the right bool values for `left_value` and `left_valid`  when the column is boolean. 
BOOL_COL  | left_value  |   left_valid
null         |   False	   |	 False
True	     |	True	   |	 True
False		 |	False	   |	 True
null		 |	False	   |	 False
False		 |	False	   |	 True
True		 |	True	   |	 True

Christian8491: For a `boolean` column, `left_value` is also boolean.
When the `value_i` is `null` or `false`, then `left_value` will be `False`.

The next column diagram illustrates the right bool values for left_value and left_valid when the column is boolean.
```
BOOL_COL | left_value | left_valid
null  | False | False
True  | True  | True
False | False | True
null  | False | False
False | False | True
True  | True  | True
```

williamBlazing: Ok, to clarify. This is the logic table as I undestand it, and  i think we agree that this is what it should be:

BOOL Col | left_value | left_valid | IS_NOT_TRUE | IS_NOT_FALSE | IS_TRUE
null            | N/A          | FALSE    | TRUE                 | TRUE                 | FALSE
TRUE       | TRUE       | TRUE     | FALSE                | TRUE                 | TRUE
FALSE      |     FALSE | TRUE     | TRUE                  | FALSE                | FALSE

This is your code:
```
if (oper == operator_type::BLZ_IS_NOT_TRUE) {
    left_valid = (left_value == false) ? true : false;
    store_data_in_buffer(static_cast<int64_t>(true), buffer, output_position);
```
Notice that you are always setting the value to true. And the valid is depending on the left_value. 
Instead it should be:
```
if (oper == operator_type::BLZ_IS_NOT_TRUE) {
     bool val = (left_valid == true && left_value == true) ? false : true;
    left_valid = true;
    store_data_in_buffer(static_cast<int64_t>(val), buffer, output_position);
```
`BLZ_IS_NOT_FALSE` also needs to be fixed

PR #1464, opened 2021-04-12
[REVIEW] Better Support for unsigned types in C++ layer

### LINKED ISSUE #1451: Unsigned Types support
Currently we do not support unsigned integer types in blazingsql. Now that they are supported in cudf we can easily add support for unsigned types in terms of compute but there are many places in the engine where we are iterating through the types.

[Calcite expression Parsing](https://github.com/BlazingDB/blazingsql/blob/9438e474f8a6b817a04db2ae41a6e0b2315cf0f3/engine/src/CalciteExpressionParsing.cpp)

[Interops](https://github.com/BlazingDB/blazingsql/tree/9438e474f8a6b817a04db2ae41a6e0b2315cf0f3/engine/src/Interpreter) 
The only real issue here is we might have to treat unsigned 64bit integers as special for certain operators because of how data is stored during execution. This can be handled pretty easily during the read phase of interops.

[Metadata Common](https://github.com/BlazingDB/blazingsql/blob/9438e474f8a6b817a04db2ae41a6e0b2315cf0f3/engine/src/io/data_parser/metadata/common_metadata.cpp#L53)

[Expression Utils](https://github.com/BlazingDB/blazingsql/blob/7a523a11c1f946aeae89e61d32189676eca3acca/engine/src/parser/expression_utils.cpp)

These are the main places where this needs to be implemented in order for it to work.

Christian8491:
This PR closes #1451


GPUtester:

engine/src/CalciteExpressionParsing.cpp
```diff
@@ -53,31 +53,58 @@ std::unique_ptr<cudf::scalar> get_max_integer_scalar(cudf::data_type type) {
...
 		using T = int64_t;
 		using ScalarType = cudf::scalar_type_t<T>;
 		static_cast<ScalarType *>(ret.get())->set_value(static_cast<T>(LONG_MAX));
 		return ret;
 	}
+	if(type.id() == cudf::type_id::UINT64) {
+		auto ret = cudf::make_numeric_scalar(type);
+		using T = uint64_t;
+		using ScalarType = cudf::scalar_type_t<T>;
+		static_cast<ScalarType *>(ret.get())->set_value(static_cast<T>(LLONG_MAX));
```
felipeblazing, line 102: im not sure this is the right max value here

felipeblazing (changes requested):

felipeblazing (changes requested):
can you add a few tests like select int + unsigned int and select unsigned int + unsigned int

Christian8491:
rerun tests

Christian8491:
rerun tests

Issue #1417, opened 2021-03-23
Support predicate push down for data providers


aucahuasi:
Related to https://github.com/BlazingDB/blazingsql/issues/1395

Currently the provider API only allow us to select which columns we want to fetch, however a much flexible design would allow to allow the providers to apply a filter at the moment they are fetching data. This is specially important for sql providers!

aucahuasi:
These points can help us to define a better strategy for the plan (when the predicate push down was already applied to the data):

1. The logical filter step also uses interops, so I think I need to tell to the filter step to do not apply the filters if these were already solved by the provider (through the predicate push down)
2. Maybe we don’t need to change the logic but instead just make more flexible that part of the engine so we support this feature... in the future we will have predicate push down for most of the providers (not only the sql ones)
3. Additionally a good thing to let the plan without changes is that most of the them doesn’t have filter steps but only projections and scans ... so indeed if we change the logic it would be for edge cases

cc @rommelDB @felipeblazing @williamBlazing

williamBlazing:
A couple clarifications:
There is a LogicalProject and a LogicalFilter which are relational algebra steps
There are also a projections and filter clause in the BindableTableScan relational algebra step.
LogicalProject and LogicalFilter  both use interops.
The filter step in the BindableTableScan also uses interops and usually behaves just like the LogicalFilter relational algebra step.
The projections clause in BindableTableScan just tells the provider/parser what columns to fetch.

For the purposes of a SQL provider, we want the SQL provider to perform the filter from the BindableTableScan because then there is less data to parse, which is an expensive step. And if we made the SQL provider perform the filter, then we dont want interops to do the filter as well. If we did, then we would be doing a copy and an operation that would be totally unnecessary.

We do not want to change anything about the physical plan. The SQL providers have nothing to do with a LogicalFilter relational algebra step, nor a LogicalProject.

williamBlazing:
What we want to do is to make the filter step of the BindableTableScan be skipped in a way that is generalizable and not necessarily have to check if its a SQL provider. 

I would suggest that we have the provider have a more general function like `filtered()` that returns a bool (true if it already provided a filter). This way its a property of the provider, and not necessarily have anything to do with SQL providers. Ultimately we would really need to look at the code to see what makes the most sense

felipeblazing:
[...]

PR #1472, opened 2021-04-14
[REVIEW] Implement predicate pushdown for data providers

aucahuasi:
https://github.com/BlazingDB/blazingsql/issues/1417


GPUtester:

engine/src/io/data_provider/sql/AbstractSQLDataProvider.cpp
```diff
@@ -53,11 +151,20 @@ std::string abstractsql_data_provider::build_select_from() const {
...
-  return "SELECT " + cols + "FROM " + this->sql.table;
-}
 
-std::string abstractsql_data_provider::build_limit_offset(size_t offset) const {
-  return " LIMIT " + std::to_string(this->sql.table_batch_size) + " OFFSET " + std::to_string(offset);
+  const size_t offset = this->sql.table_batch_size * (this->total_number_of_nodes * batch_index + this->self_node_idx);
+  std::string limit = " LIMIT " + std::to_string(this->sql.table_batch_size) + " OFFSET " + std::to_string(offset);
+  auto ret = "SELECT " + cols + "FROM " + this->sql.table;
+
+  if (sql.table_filter.empty()) {
```
williamBlazing, line 159: if your table_filter is not empty and the `where` is not empty you can do:
`" where (" + sql.table_filter + ") AND (" + this->where + ")"`

williamBlazing (changes requested):
Is this PR also implementing PostgreSQL support or just predicate pushdown?
Please add unit tests for the predicate parsing or transpiling functions, there is alot of logic there, and it should be really easy to setup unit tests for that.

aucahuasi:
> Is this PR also implementing PostgreSQL support or just predicate pushdown?
> Please add unit tests for the predicate parsing or transpiling functions, there is alot of logic there, and it should be really easy to setup unit tests for that.

@williamBlazing PostgreSQL and SQLite support is another PR, this PR just set the common interface between those backends
cc @gcca

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
```diff
@@ -278,7 +309,7 @@ ral::execution::task_result BindableTableScan::do_process(std::vector< std::uniq
     std::unique_ptr<ral::frame::BlazingTable> filtered_input;
 
     try{
-        if(this->filtered) {
+        if(this->filtered && !this->predicate_pushdown_done) {
```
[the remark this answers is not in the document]
aucahuasi: ok

engine/src/io/data_provider/sql/AbstractSQLDataProvider.cpp
aucahuasi: Good idea!

aucahuasi:
>Please add unit tests for the predicate parsing or transpiling functions, there is alot of logic there, and it should be really easy to >setup unit tests for that.

sure, I'll add some unit tests!

engine/src/execution_graph/logic_controllers/BatchProcessing.cpp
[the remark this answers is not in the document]
aucahuasi: done

aucahuasi:
The sql transpiler has unit tests now! Done @williamBlazing

engine/src/io/data_provider/sql/AbstractSQLDataProvider.cpp
aucahuasi: Done ty

felipeblazing (approved):

williamBlazing:
rerun-tests

engine/tests/provider/sql_transpiler_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+    "BindableTableScan(table=[[main, lineitem]], filters=[[<(+($5, $6), 100)]])",
+    "l_extendedprice + l_discount < 100"
+  ),
+  SQLTranspilerTestParam(
+    "BindableTableScan(table=[[main, lineitem]], filters=[[<(+($5, $6), 100)]])",
+    "((l_extendedprice + l_discount) < 100)",
+    sql_tools::get_default_operators()
+  ),
+  SQLTranspilerTestParam(
+    "BindableTableScan(table=[[main, lineitem_MYSQL]], filters=[[AND(>=($10, 1995-01-01), <=($11, 1996-12-31))]], projects=[[0, 2, 5, 6, 10]], aliases=[[l_orderkey, l_suppkey, l_extendedprice, l_discount, l_shipdate]])",
```
williamBlazing, line 97: I think this is incorrect. `projects=[[0, 2, 5, 6, 10]]`  happens first, which means that the columns referred to in filters can only be $0 - $5, but you have $10 and $11.

engine/tests/provider/sql_transpiler_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+    "((l_shipdate >= \"1995-01-01\") AND (l_commitdate <= \"1996-12-31\"))",
+    sql_tools::get_default_operators()
+  ),
+  SQLTranspilerTestParam(
+    "BindableTableScan(table=[[main, lineitem_MYSQL]], filters=[[AND(<($6, $5), IS NOT NULL($0))]], projects=[[0, 11, 12]], aliases=[[l_orderkey, l_commitdate, l_receiptdate]])",
```
williamBlazing, line 102: Same issue here. How is this working? Is it because the transpiler receives a set of column names and it does not know about that was in the project clause? And this example is simply setup for the unit test, and not necessarily "real"

williamBlazing: If the relational algebra is not "Real"  then i think we should make it real, and have the test could take what is in the project, to then properly obtain what should be set in the column names and column indexes and we can have proper relational algebra in the test. This would make it look more real and also if someone looks at the unit test to better understand the transpiler, then they would know that the column names and indexes it needs to use are not those of the table schema as it is, but based off of the project

williamBlazing (changes requested):

engine/tests/provider/sql_transpiler_test.cpp
aucahuasi: Yeah, the predicate transpiler only reads the filter part, that's why is working fine, so I think for those unit test we can just change the projection values (i changed the filter variables and forgot to update the projection ones)
cc @gcca

gcca: @williamBlazing I've updated the projects and aliases fields to consider the column indices in the following commit
https://github.com/BlazingDB/blazingsql/pull/1472/commits/3f51d11852059fbcc9f8cedf5c904a689dea027b

williamBlazing (approved):

PR #1473, opened 2021-04-15
Added Warning to Window Functions

williamBlazing:
Adding a warning `WARNING: Window Functions are currently an experimental feature and not fully supported or tested` for any time a user runs a window function. This will be removed when we are more confident in the feature. Right now we know it has a lot of holes and bugs


GPUtester:

PR #1483, opened 2021-04-19
[REVIEW] Update build dependencies for branch-0.20

romulo-auccapuclla:
Changes:
1. Add dependencies script for branch-0.20
2. Update cuda support 11+ for nightly version in README
3. Update conda meta yaml to pinning nlohmann_json to version 3.9.1


GPUtester:

williamBlazing:
add to whitelist

williamBlazing:
ok to test

williamBlazing:
Something I dont like about how we have some things setup here, and honestly they were not that much better before either, is that we are having more and more dependencies for external components that are actually optional, for example support for S3, GCS, mysql, posgresql and sqllite. We have build flags to disable those which is great, but when we talk about our dependencies and our dependencies.sh script, we dont make it easy, or describe properly which of those dependencies are necessary and which go with S3, GCS, mysql, etc... 

I propose that the dependencies.sh script take in the same disabling flags that our build.sh script does and that way its all connected and its easier to make a smaller, quicker simpler build. And the dependencies.sh script with those flags would self document which dependencies are necessary for what optional feature

PR #1505, opened 2021-04-26
[REVIEW] Support for right outer join

### LINKED ISSUE #1391: Add support for right outer join
It should be relatively simple to implement a right outer join, simply by swapping the left and right sides of the tables.

Also, Apache Calcite will sometimes optimize a full outer join to a right outer join, which right now will fail and confuse the user. For example this query:

```
                    select l.l_orderkey, o.o_orderkey, l.l_partkey, o.o_custkey from lineitem as l
                    full outer join orders as o on l.l_orderkey = o.o_orderkey
                    and l.l_partkey < o.o_custkey
                    where o.o_orderdate < '1992-01-01'"""
```

Christian8491:
This PR closes #1391.  In order to this PR be successful is needed this PR BlazingDB/blazingsql-testing-files#28 will be merged first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ## New Features
 - #1445 Support for CURRENT_DATE, CURRENT_TIME and CURRENT_TIMESTAMP
+- #1505 Support for right outer join
 
 
 ## Improvements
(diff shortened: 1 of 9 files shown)


Christian8491:
rerun tests

PR #1542, opened 2021-05-18
[REVIEW] Clean code

kharoc:


GPUtester:

kharoc:
rerun tests

kharoc:
rerun tests

tests/BlazingSQLTest/Runner/runTest.py
```diff
@@ -1707,13 +1181,54 @@ def run_query_log(
...
+                    print_result,
+                    engine,
+                    input_type,
+                    load_time,
+                    engine_time,
+                    total_time,
+                    comparing
+                )
+
+        else:
```
williamBlazing, line 1228: what is `print_query_results2` ?  it can probably have a better name

tests/BlazingSQLTest/Runner/runTest.py
```diff
@@ -1707,13 +1181,54 @@ def run_query_log(
...
+                    base_results_gd.resultSet, expected_dtypes
+                )
+                pdf2 = upcast_to_float(pdf2).fillna(get_null_constants(pdf2))
+
+                formatResults(pdf1, pdf2, worder, orderBy)
+            else:
+                if compareResults == "true":
+                        format_pdf(pdf1, worder, orderBy)
+                        pdf2 = get_results(resultFile)
+                        print(pdf2)
```
williamBlazing, line 1202: should this print be there?

tests/BlazingSQLTest/Runner/runTest.py
```diff
@@ -1676,28 +1148,30 @@ def run_query(
...
-
-    try:
-        result_gdf = bc.log(query)
-    except Exception as e:
-        error_message=str(e)
+                        comparing,
+                        compareResults,
+                        resultFile)
+      
+def results_processing(result_gdf,
```
williamBlazing, line 1155: Could you please add comments to document this function. What does it do? what are all these parameters?

tests/BlazingSQLTest/Runner/runTest.py
```diff
@@ -1707,13 +1181,54 @@ def run_query_log(
...
+                        pdf2 = get_results(resultFile)
+                        print(pdf2)
+                else:
+                    pdf2 = pd.DataFrame()
+                    formatResults(pdf1, pdf2, worder, orderBy)
+
+            if Settings.execution_mode == ExecutionMode.GENERATOR:
+                results_file_generator(testsWithNulls, filename, engine, pdf2)
+            else:
+                print_query_results(
```
williamBlazing, line 1210: Could you please refactor some of this as well? Here the `print_query_results` function also does the validation of the results and query names and a whole bunch more things. The function name is misleading because it sound like this query should "print query results" but its really doing a whole lot more. I would suggest having validation and printing  being separate functions and having them be called sequentially instead of being nested, that way those functions are more reusable and modular.

williamBlazing (changes requested):

tests/BlazingSQLTest/Runner/runTest.py
kharoc: Done

kharoc: It's shoudn't be here

tests/BlazingSQLTest/Runner/runTest.py
```diff
@@ -322,22 +325,70 @@ def print_fixed_log(
     logger.info(total_time)
 
 
-def print_query_results(
+def comparing_results(  pdf1,
```
williamBlazing, line 328: this `comparing_results` and the above `compare_results` have very very similar names and it can be confusing. I propose we call this one `compare_test_results` and the above `compare_result_values` since this one is about all the results while the one above is only looking at the actual values

tests/BlazingSQLTest/Runner/runTest.py
```diff
@@ -1052,6 +692,13 @@ def on_jenkins():
 
 
 def print_tests(tests, onlyFails=False):
+    """
+        After we have the pdf log with this function we are printing the values on the screem at the end of the execution tests.
```
williamBlazing, line 696: typo `screem` => `screen`

williamBlazing (changes requested):

kharoc:
rerun tests

williamBlazing:
rerun tests

kharoc:
rerun tests

PR #1552, opened 2021-05-25
VDE-181 [REVIEW] Feature/concurrent e2e

diegodfrf:


GPUtester:

tests/BlazingSQLTest/Runner/config.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+RUN_SETTINGS:
+#  SAVE_LOG: "false" - Not yet supported in yaml.
+  EXEC_MODE: gpuci #gpuci, full, generator
+  NGPUS: 1
+  NETWORK_INTERFACE: lo
+  WORKSHEET: BSQL Log Results # or "BSQL Performance Results"
+  LOG_INFO:
+  CONCURRENT: False
+#  CONCURRENT:  - Not yet supported in yaml.
```
williamBlazing, line 17: not supported? but the line above looks like it is?

williamBlazing (commented):
Can you explain a little about what this PR will provide us with. 
I understand it enables concurrent query tests. When do those get run? Is it one specific test set that is run concurrently? is it all of them?

tests/BlazingSQLTest/Runner/config.yaml
diegodfrf: The plan is to support multiple configurations at the same time. At the moment only one setting works for that parameter. For example, for the following configuration:
``` yaml
CONCURRENT:
   - False
   - True 
```
Then you will run two sets of tests, one for `CONCURRENT = True` and one for `CONCURRENT = False`, and thus simplify `test.sh` at this point. This part not yet support.

diegodfrf:
This PR provides us with the following:
* **Configuration file creation config.yaml**: Global settings will load from here and other options were added.
``` yaml
# config.yaml
RUN_SETTINGS:
  CONCURRENT: False
SETUP:
  APPLY_ORDER: True
  USE_PERCENTAGE: False
  ACCEPTABLE_DIFFERENCE: 0.01
  PRINT_RESULT: True
  COMPARE_WITH: drill
  COMPARING: True
  DATA_TYPES:
    - CUDF
    - DASK_CUDF
    - ORC
    - PARQUET
    - CSV
    - JSON
```
* **Support Concurrent E2E test**: All E2E test are support, except for legacyTest. This is configured from config.yaml
``` yaml
# config.yaml
RUN_SETTINGS:
  CONCURRENT: False
```

* **Support skip_with for concurrent**: If a SuiteTest needs to be skipped for concurrency, it can be disabled.

This option skip test when concurrent is True
``` yaml
# config.yaml
TEST_SUITE:
  SETUP:
    SKIP_WITH:
      - CONCURRENT: True
```
* **A configuration file can be specified using the --config-file parameter**:  The same option will be added for test.sh
``` bash
python -m EndToEndTests.mainE2ETests --config-file config.yaml
```
* **Some E2E environment variables were migrated to config.yaml**: The goal is to migrate all or most. The following environment variables are here now
``` yaml
# config.yaml
TEST_SETTINGS:
  LOG_DIRECTORY: $CONDA_PREFIX/
  DATA_SIZE: "100MB2Part"
  EXECUTION_ENV: "local"
  DASK_CONNECTION: "local"

RUN_SETTINGS:
  EXEC_MODE: gpuci
  NGPUS: 1
  NETWORK_INTERFACE: lo
  WORKSHEET: BSQL Log Results
  LOG_INFO:
  CONCURRENT: False
```

diegodfrf:
Only TPCHSuite will execute for concurrent True and False

PR #1559, opened 2021-06-03
Update ucx-py versions

ajschmidt8:
This PR updates the `ucx-py` versions used by Blazing. `ucx-py` will not be switching to calendar versioning like the rest of RAPIDS, so their versions must be updated accordingly (similar to https://github.com/rapidsai/cuml/pull/3911 and https://github.com/rapidsai/cugraph/pull/1634).

For now I've hard-coded the values here to help unblock some RAPIDS builds, but I will update Blazing accordingly if we set up any automation here.


GPUtester:

PR #1572, opened 2021-06-17
[REVIEW] setting up default value for max_bytes_chunk_read to 256 MB

### LINKED ISSUE #1462: Default max_bytes_read for csv when creating table
when creating a csv table the user should explicitly set max_bytes_chunk_read to none if the user does not want to set this value. Otherwise the default value should be 268435456 bytes.

Christian8491:
This PR closes #1462


GPUtester:

Christian8491:
rerun tests

pyblazing/pyblazing/apiv2/context.py
```diff
@@ -2479,17 +2479,19 @@ def create_table(self, table_name, input, **kwargs):
                 parsedMetadata = parseHiveMetadata(table, uri_values)
                 table.metadata = parsedMetadata
 
-            has_csv_metadata = False
+            # When reading CSV files, we want to set up by default the size of chunks
             if "max_bytes_chunk_read" in parsedSchema["args"].keys():
-                if parsedSchema["args"]["max_bytes_chunk_read"] > 0:
-                    has_csv_metadata = True
+                if parsedSchema["args"]["max_bytes_chunk_read"] <= 0:
+                    parsedSchema["args"]["max_bytes_chunk_read"] = 268435456  # 256 MBs
```
wmalpica, line 2485: This value, max_bytes_chunk_read,  should be a config_option

wmalpica (changes requested):

pyblazing/pyblazing/apiv2/context.py
Christian8491: As far as I know, we only use this parameter for the CSV reader. I'm not sure if we can currently use this for other readers as well. If it's used for CSV only, I would recommend leaving this parameter as is now. So this will only be used when the `create_table()` function is called and not for the `BlazingContext()`.

wmalpica: ok, fair point

PR #1573, opened 2021-06-17
[WIP] fix current_date inside filter

gcca:


GPUtester:

wmalpica:
We previously implemented support for CURRENT_DATE, CURRENT_TIME and CURRENT_TIMESTAMP. It seems like those were previously only implemented for a Project, but not able to be used in a transformation.
One of the things we realized in the early implementation is that you want to make sure that the CURRENT_TIMESTAMP (or similar) is the same for the duration of the whole query and for the whole cluster, otherwise you can end up with some funky behavior. 
Please take a look at the previous implementation and use some of the similar concepts. 
In interops, we should not be obtaining the current date or time for every calculation. Instead, when the plan for interops evaluation is made, which should convert the CURRENT_X to a literal value based on the current time that is stored in the context.
Additionally, lets make sure that a predicate for CURRENT_DATE, CURRENT_TIME and CURRENT_TIMESTAMP all work, not just CURRENT_DATE.

PR #1597, opened 2021-08-19
Restrict Google cloud cpp version

raydouglass:
Starting with `google-cloud-cpp>=1.30`, it requires `libgcc-ng>=9.4.0` which is not compatible with RAPIDS. RAPIDS requires `libgcc-ng==9.3.0`.

I've testing this build and install with RAPIDS `21.08` locally and it works with this change.


raydouglass:
rerun tests

raydouglass:
rerun tests

wmalpica:
rerun tests

PR #1600, opened 2021-09-29
Fix double free

jglaser:
Somehow this fix for a UCX double free seg fault only made it into branch-0.19, but was never taken up into the current working branch.


wmalpica:
Looking into the history from what i can tell is that we released 0.19, then did these changes, and then released 0.19.1, but those changes never made it into 0.20 and so on. BUT they are in the main branch, so they should be in all the conda packages for all stable releases.

